PR #184, opened 2023-11-29
chore: PC-10133 Move AlertPolicy to separate package

daniel-zelazny:
This PR:

- moves AlertPolicy to separate package and rewrites its validation to use validaiton library.
- removes PublicAlertMethod
- add temporary AlertMethodRef package allowing to return AlertMethod details with AlertPolicy in old manner.
- introduces fix to assert method allowing to look for error match for more than one error
- moves Severity and Measurement object definitions to alertpolicy package because they are used only with this particular context


nieomylnieja (changes requested):

nieomylnieja (changes requested):

nieomylnieja (changes requested):

nieomylnieja (approved):

nieomylnieja (changes requested):

nieomylnieja (approved):

navayla777-dotcom (commented):
Altert policy moving #49  dashboard platform

PR #227, opened 2024-01-04
Revert postUpgradeTasks from Renovate config

nieomylnieja:
It turns out these are only enabled for self hosted Renovate instances...


daniel-zelazny (approved):
:+1:

PR #235, opened 2024-01-09
Create release.yml

nieomylnieja:
## Motivation

To make it easier to group changes between versions into meaningful categories we're adding `release.yml` template.


skrolikiewicz (approved):
🚀

PR #238, opened 2024-01-12
feat: sumologic timeslice validation [PC-7427]

nikodemrafalski:
Tightens sumo logic log query validation to allow timeslice operator to be used with 15,30 and 60 sec as an argument

### DIFF
--- a/manifest/v1alpha/slo/metrics_sumo_logic.go
+++ b/manifest/v1alpha/slo/metrics_sumo_logic.go
@@ -145,18 +145,17 @@ func validateSumoLogicTimeslice(query string) error {
 	return nil
 }
 
-func getTimeSliceFromSumoLogicQuery(query string) (time.Duration, string, bool, error) {
+func getTimeSliceFromSumoLogicQuery(query string) (
+	tsDuration time.Duration, durationString string, containsAlias bool, err error,
+) {
 	r := regexp.MustCompile(`\stimeslice\s([-+]?(\d+[a-z]+\s?)+)\s(?:as n9_time)?`)
 	matchResults := r.FindAllStringSubmatch(query, 2)
-
 	if len(matchResults) == 0 {
 		return 0, "", false, fmt.Errorf("query must contain a 'timeslice' operator")
 	}
-
 	if len(matchResults) > 1 {
 		return 0, "", false, fmt.Errorf("exactly one 'timeslice' declaration is required in the query")
 	}
-
 	submatches := matchResults[0]
 
 	if submatches[1] != submatches[2] {
@@ -164,12 +163,12 @@ func getTimeSliceFromSumoLogicQuery(query string) (time.Duration, string, bool, 
 	}
 
 	// https://help.sumologic.com/05Search/Search-Query-Language/Search-Operators/timeslice#syntax
-	durationString := strings.TrimSpace(submatches[1])
-	containsAlias := strings.Contains(submatches[0][1:], "as n9_time")
-	timeslice, err := time.ParseDuration(durationString)
+	durationString = strings.TrimSpace(submatches[1])
+	containsAlias = strings.Contains(submatches[0][1:], "as n9_time")
+	tsDuration, err = time.ParseDuration(durationString)
 	if err != nil {
 		return 0, "", containsAlias, fmt.Errorf("error parsing timeslice duration: %s", err.Error())
 	}
 
-	return timeslice, durationString, containsAlias, nil
+	return tsDuration, durationString, containsAlias, nil
 }
--- a/manifest/v1alpha/slo/metrics_sumo_logic.go
+++ b/manifest/v1alpha/slo/metrics_sumo_logic.go
@@ -45,15 +45,15 @@ var sumoLogicCountMetricsLevelValidation = validation.New[CountMetricsSpec](
 				if *good.Type != "logs" || *total.Type != "logs" {
 					return nil
 				}
-				goodTS, _, _, err := getTimeSliceFromSumoLogicQuery(*good.Query)
+				goodTimeSlice, err := getTimeSliceFromSumoLogicQuery(*good.Query)
 				if err != nil {
 					return nil
 				}
-				totalTS, _, _, err := getTimeSliceFromSumoLogicQuery(*total.Query)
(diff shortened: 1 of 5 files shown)


manifest/v1alpha/slo/metrics_sumo_logic_test.go
```diff
@@ -187,22 +187,78 @@ _collector="n9-dev-tooling-cluster" _source="logs"
   | sort by n9_time asc`,
 			Error: testutils.ExpectedError{
 				Prop:    "spec.objectives[0].rawMetric.query.sumoLogic.query",
-				Message: "exactly one timeslice declaration is required in the query",
+				Message: "query must contain a 'timeslice' operator",
```
bdw-nobl9, line 190: [nit] Do we want to put this message in a `const` and reuse it?

It's also used here:
https://github.com/nobl9/nobl9-go/pull/238/files#diff-77a902bb39ea318908cd500a58d5cf1454e6229f90b04fb53314b1d9e89e0eceR154

manifest/v1alpha/slo/metrics_sumo_logic.go
```diff
@@ -136,21 +125,50 @@ var sumoLogicLogsTypeValidation = validation.New[SumoLogicMetric](
...
+	}
+	submatches := matchResults[0]
+
+	if submatches[1] != submatches[2] {
+		return 0, "", false, fmt.Errorf("timeslice interval must be in a NumberUnit form - for example '30s'")
+	}
+
 	// https://help.sumologic.com/05Search/Search-Query-Language/Search-Operators/timeslice#syntax
-	timeslice, err := time.ParseDuration(subMatch[1])
+	durationString = strings.TrimSpace(submatches[1])
```
bdw-nobl9, line 166: [nit] We could, for example, create a map using enums for every part (`duration`, `contains`, and every other part) instead of using _magic_ numbers [forcing to do a mental map that `1` is _duration_, `2` is _contains_, etc.] - but it's only a nitpicked idea :peace_symbol:

manifest/v1alpha/slo/metrics_sumo_logic.go
```diff
@@ -136,21 +125,50 @@ var sumoLogicLogsTypeValidation = validation.New[SumoLogicMetric](
...
+func getTimeSliceFromSumoLogicQuery(query string) (
+	tsDuration time.Duration, durationString string, containsAlias bool, err error,
+) {
+	r := regexp.MustCompile(`\stimeslice\s([-+]?(\d+[a-z]+\s?)+)\s(?:as n9_time)?`)
+	matchResults := r.FindAllStringSubmatch(query, 2)
+	if len(matchResults) == 0 {
+		return 0, "", false, fmt.Errorf("query must contain a 'timeslice' operator")
+	}
+	if len(matchResults) > 1 {
+		return 0, "", false, fmt.Errorf("exactly one 'timeslice' declaration is required in the query")
```
bdw-nobl9, line 157: Wondering about words used in this message: `declaration` vs  `usage` :thinking: 
I connect declaring something with declaring a variable.
But it's nothing dealbreaking (-; [nit] :dove:

manifest/v1alpha/slo/metrics_sumo_logic_test.go
```diff
@@ -187,22 +187,78 @@ _collector="n9-dev-tooling-cluster" _source="logs"
...
+  | json "log"
+  | timeslice 15000ms as n9_time
   | parse "level=* *" as (log_level, tail)
   | if (log_level matches "error" ,0,1) as log_level_not_error
   | sum(log_level_not_error) as n9_value by n9_time
   | sort by n9_time asc`,
 			Error: testutils.ExpectedError{
 				Prop:    "spec.objectives[0].rawMetric.query.sumoLogic.query",
-				Message: "exactly one timeslice declaration is required in the query",
+				Message: "timeslice value must be 15, 30, or 60 seconds, got: [15000ms]",
```
bdw-nobl9, line 261: [nit] Same thing about making part of this message a const.

bdw-nobl9 (approved):
Leaving loose ideas and nitpicks, but approved! :seal:

manifest/v1alpha/slo/metrics_sumo_logic.go
```diff
@@ -136,21 +125,50 @@ var sumoLogicLogsTypeValidation = validation.New[SumoLogicMetric](
...
-		return 0, fmt.Errorf("timeslice declaration must match regular expression: %s", sumoLogicTimeSliceRegexp)
+
+	if !containsAlias {
+		return errors.New("timeslice operator requires an n9_time alias")
 	}
+	return nil
+}
+
+func getTimeSliceFromSumoLogicQuery(query string) (
+	tsDuration time.Duration, durationString string, containsAlias bool, err error,
```
nieomylnieja, line 149: maybe consider returning a struct for the 3 first values? :)

bdw-nobl9:
> @Bartek @nieomylnieja  
> 
> 
> 
> I'm not a fan of adding a const just because a particular string is used in code and in tests.
> 
> If I introduce a const with an error message and rely on it in tests, someone can change the value of the string in const, and the test will still pass. 
> 
> 
> 
> I would do it if the value were used twice in the actual system under test, but it's used once in this case.
> 
> 
> 
> 

Fine for me (-: especially it would have only a single real usage

bdw-nobl9:
> @Bartek @nieomylnieja  
> 
> 
> 
> I'm not a fan of adding a const just because a particular string is used in code and in tests.
> 
> If I introduce a const with an error message and rely on it in tests, someone can change the value of the string in const, and the test will still pass. 
> 
> 
> 
> I would do it if the value were used twice in the actual system under test, but it's used once in this case.
> 
> 
> 
> 

You mentioned some other Bartek (-;

nieomylnieja:
> > @bartek @nieomylnieja
> > I'm not a fan of adding a const just because a particular string is used in code and in tests.
> > If I introduce a const with an error message and rely on it in tests, someone can change the value of the string in const, and the test will still pass.
> > I would do it if the value were used twice in the actual system under test, but it's used once in this case.
> 
> You mentioned some other Bartek (-;

You're right :D sorry for that :bow:

PR #257, opened 2024-02-03
feat: Switch back to native yaml decoding

nieomylnieja:
## Motivation

When recently working on Data Export, we've encountered a problem using `json.RawMessage` with `go-yaml` decoding.
The raw message was used to postpone decoding of Data Export spec until we could decide what type should it be decoded to.
Due to time limitations back then we went with the easier solution and decided to use `yaml.YAMLToJSON`, which converts YAML to JSON and then uses `encoding/json` to decode the spec.

This is not ideal.
Doing it this way we're adding a decoding overhead, that's one thing.
Another, more serious issue is that we're stripping ourselves from receiving detailed syntax errors for YAML.
Instead, we're always getting JSON errors.

By adding `serdeutil` package and `RawMessage` which implements both JSON and YAML (un)marshalers we're bringing the aforementioned capabilities back home.

## Release Notes

Syntax errors when parsing YAML now return more specific details.


nieomylnieja:
As a side note, I can't really recall why this was a problem in first place:

```go
		// Workaround for https://github.com/goccy/go-yaml/issues/313.
		// If the library changes its interpretation of empty pointer fields,
		// we should switch to native yaml.Unmarshal instead.
```

All the unit tests are passing, and revisiting the mentioned issue actually makes me wonder If it is in fact an issue :shrug:

internal/serdeutil/raw_message.go
```diff
@@ -0,0 +1,31 @@
+package serdeutil
```
daniel-zelazny, line 1: Could you explain why it is named like this? It is the first time I see that phrase :)

internal/serdeutil/raw_message.go
```diff
@@ -0,0 +1,31 @@
...
+func (m RawMessage) MarshalYAML() ([]byte, error) { return m.marshal() }
+
+func (m RawMessage) marshal() ([]byte, error) {
+	if m == nil {
+		return []byte("null"), nil
+	}
+	return m, nil
+}
+
+func (m *RawMessage) UnmarshalJSON(data []byte) error { return m.unmarshal(data) }
```
daniel-zelazny, line 22: Go Docs recommends using consistent receivers definition, it there any reason you have used value in marshal operation?

internal/serdeutil/raw_message.go
nieomylnieja: serde is short for serialization/deserialization, I know of Java and Rust which use this nomenclature

internal/serdeutil/raw_message.go
nieomylnieja: it's a limitation of json encoding library, if I have a struct which is included as a field value of another struct and I don't want it to be a pointer It won't get marshaled, see https://github.com/golang/go/issues/22967 
The ongoing discussion on json v2 library also describes this behaviour: https://github.com/golang/go/discussions/63397

PR #261, opened 2024-02-06
feat: Add fields to Agent Status showing newest available agent versions [PC-11737]

marcinlawnik:
## Motivation

We want to expose this information through the API, besides release notes and DockerHub/image repositories.

## Summary

Added two fields to `Agent` status struct (as this information is dynamic and not a part of `Agent` definition.)


altar64 (approved):
👍🏼

PR #262, opened 2024-02-06
feat: PC-11843 budget adjustment kind

kubaceg:
## Summary

Added new kind BudgetAdjustment for handling budget adjustment events.

## Release Notes

New `BudgetAdjustment` kind which allows customers to schedule recurring future events that should be excluded from the Error Budget calculations in their SLOs


nieomylnieja:
Please add Release Notes section with a short description of this new object we're adding and what it does :)

manifest/v1alpha/budgetadjustment/validation.go
```diff
@@ -0,0 +1,89 @@
...
+		Required(),
+	validation.For(func(s Slo) string { return s.Name }).
+		WithName("name").
+		Required(),
+)
+
+func durationValidationRule() validation.SingleRule[time.Duration] {
+	return validation.NewSingleRule(func(v time.Duration) error {
+		if v.Truncate(time.Minute) != v {
+			return errors.New("duration must be in whole minutes without seconds")
```
nieomylnieja, line 65:
```suggestion
			return errors.New("duration must be defined with minutes precision")
```

manifest/v1alpha/budgetadjustment/validation.go
```diff
@@ -0,0 +1,89 @@
...
+
+func rruleValidationRule() validation.SingleRule[string] {
+	return validation.NewSingleRule(func(v string) error {
+		if len(v) == 0 {
+			return nil
+		}
+
+		_, err := rrule.StrToRRule(v)
+		if err != nil {
+			return errors.Wrap(err, "invalid rrule")
```
nieomylnieja, line 80: this error message doesn't bring any value, `validation` pkg includes the property name in the final error, so we know it's `rrule` and obviously it's invalid :P

```suggestion
			return err
```

manifest/v1alpha/budgetadjustment/validation.go
```diff
@@ -0,0 +1,89 @@
...
+	validation.For(func(s Spec) time.Time { return s.FirstEventStart }).
+		WithName("firstEventStart").
+		Required(),
+	validation.For(func(s Spec) time.Duration { return s.Duration }).
+		WithName("duration").
+		Required().
+		Rules(durationValidationRule()),
+	validation.For(func(s Spec) string { return s.Rrule }).
+		WithName("rrule").
+		Rules(rruleValidationRule()),
```
nieomylnieja, line 40: you can use `validation.Transform` to ensure this string is of type `rrule.Rule`:

```suggestion
	validation.Transform(func(s Spec) string { return s.Rrule }, rrule.StrToRRule).
		WithName("rrule"),
```

manifest/v1alpha/budgetadjustment/validation.go
```diff
@@ -0,0 +1,89 @@
...
+		IncludeForEach(sloValidationRule),
+)
+
+var sloValidationRule = validation.New[Slo](
+	validation.For(func(s Slo) string { return s.Project }).
+		WithName("project").
+		Required(),
+	validation.For(func(s Slo) string { return s.Name }).
+		WithName("name").
+		Required(),
```
nieomylnieja, line 59: both name and project should adhere to our DNS name constraints

manifest/v1alpha/budgetadjustment/budget_adjustment.go
```diff
@@ -0,0 +1,52 @@
...
+type Spec struct {
+	Description     string        `json:"description,omitempty"`
+	FirstEventStart time.Time     `json:"firstEventStart"`
+	Duration        time.Duration `json:"duration"`
+	Rrule           string        `json:"rrule,omitempty"`
+	Filters         Filters       `json:"filters"`
+}
+
+type Filters struct {
+	Slos []Slo `json:"slos"`
```
nieomylnieja, line 46:
```suggestion
	SLOs []Slo `json:"slos"`
```

manifest/v1alpha/budgetadjustment/budget_adjustment.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type Slo struct {
```
nieomylnieja, line 49:
```suggestion
type SLO struct {
```

nieomylnieja (changes requested):

manifest/v1alpha/budgetadjustment/validation.go
kubaceg: You're right

jakubpieta (approved):
🚀

manifest/v1alpha/budgetadjustment/budget_adjustment.go
```diff
@@ -43,10 +43,10 @@ type Spec struct {
 }
 
 type Filters struct {
-	Slos []Slo `json:"slos"`
+	SLOs []SLO `json:"slos"`
 }
 
-type Slo struct {
+type SLO struct {
```
nieomylnieja, line 49: just one final thing, I think we should name it `SLORef`, similar to how we name such references in Role Binding and Alert Policy

nieomylnieja (approved):
looks good!

manifest/v1alpha/budgetadjustment/budget_adjustment.go
nieomylnieja: the tag for `slos` can stay as is, this is more of a code thing, when using completion it will be easier to have a single result for SLO instead of two

nieomylnieja (approved):

PR #288, opened 2024-02-21
feat: Switch to new major JWT library and new JWK handler

nieomylnieja:
## Motivation

Switching from [jwt](https://github.com/golang-jwt/jwt) v4 to v5 and from [jwx](https://github.com/lestrrat-go/jwx) v1 to v2 has been a long due task.
Since the contracts for both libraries have changed, the transition had to be done with care and manual modifications.

## Summary

jwt library has been bumped from v4 to v5.
There were only minor changes its API.

Since jwx library had a major refactor of its API, I began wondering If it's not a good time to switch to something lighter.
jwx is a heavy and robust project, with many dependencies and functionalities which far outstretch our needs.

Upon some web foraging I decided to give [keyfunc](https://github.com/MicahParks/keyfunc) a chance.
The library focuses on exactly one thing, providing key function to `jwt.Parse` method.
It's underlying, sole dependency, [jwkset](https://github.com/MicahParks/jwkset) is a library (no bloated dependencies) which still provides us with the means to run holistic tests with artificial JWK authority.

Along with removing jwx, I managed to also remove 3 other dependencies:
- zerolog
- go-cache
- mapstructure

I also uncovered that for preset access token (for example, one read from config.toml), SDK was still running a request to fetch a new token, even if the current token was not expired yet :sweat:
This was also fixed.

## Testing

Unit tests have been adapted to the change.

Minimal program:

```go
package main

import (
	"context"
	"encoding/json"
	"log"
	"os"

	"github.com/nobl9/nobl9-go/sdk"
	v1 "github.com/nobl9/nobl9-go/sdk/endpoints/objects/v1"
)

func main() {
	client, err := sdk.DefaultClient()
	if err != nil {
		log.Fatal(err)
	}

	projects, err := client.Objects().V1().GetV1alphaProjects(context.Background(), v1.GetProjectsRequest{
		Names: []string{"default"},
	})
	if err != nil {
		log.Fatal(err)
	}

	enc := json.NewEncoder(os.Stdout)
	enc.SetIndent("", "  ")
	_ = enc.Encode(projects)
}
```

## Release Notes

Underlying JWT handler chain has been changed. This should not influence how the existing or new programs run in general. However, specific errors returned by SDK when there's a problem with parsing or verifying the token have been changed.


sdk/jwt.go
```diff
@@ -3,197 +3,207 @@ package sdk
...
-	if err != nil {
-		return nil, err
+func newJWTParser(issuer, jwkFetchURL string) *jwtParser {
+	return &jwtParser{
+		parser: jwt.NewParser(
+			jwt.WithValidMethods([]string{jwtSigningAlgorithm.Alg()}),
+			// Applies to "exp", "nbf" and "iat" claims.
+			jwt.WithLeeway(jwtLeeway),
+			jwt.WithExpirationRequired(),
+			// "exp" amd "nbf" claims are always verified, "iat" is optional as per JWT RFC.
```
daniel-zelazny, line 134:
```suggestion
			// "exp" and "nbf" claims are always verified, "iat" is optional as per JWT RFC.
```

;)

daniel-zelazny (approved):
Amazing job :open_mouth:, I very appreciate your tendency to look for alternatives, is there anything against introducing github.com/MicahParks/jwkset to n9?

PR #291, opened 2024-02-23
doc: Document release and development workflows

nieomylnieja:
## Motivation

The purpose of this PR is to document both our release process and development workflow.
These documents will most likely be adjusted over time and the rules are not set in stone.
However, we need some written guidelines and over the past two months since SDK public release, we've already had a fair share of scenarios which helped define these flows.

## Summary

- Added `RELEASE.md` for release process and `DEVELOPMENT.md`.
- The documents were placed under `docs` to avoid further cluttering the root folder.
- Extended markdownlint formatting to other readme's in the repo.
- Added markdownlint config file with better tuned options for our use cases.
- Formatted other, previously not linted readme's in the repo.
- Updated `release-notes.sh` to better handle scenarios when either `---` co-authors notice was included in commit message and exclude Renovate PRs from `## Release Notes` header detection.


BSski (approved):
I like it a lot :)

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
+# Release process
+
+The release process for nobl9-go cannot be discussed without the context of
+our platform releases. In the following document, `n9` will refer to our
```
pawel-nobl9, line 4: avoid using words like our, us, make it more formal and precise, i.e. say 'the context of Nobl9 platform releases' instead

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+platform **production** version.
+
+## The process
+
+1. Create release candidate version. \
```
pawel-nobl9, line 9: imho it would be good to highlight this line (and also the 2nd. point line so it will be visually easier to see that process consist of those two points,
you can use ### to do that and than treat description as a paragraph describing the point

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+  Most likely this step is where you can stop.
+  Creating an official version release is only necessary if you want our users
```
pawel-nobl9, line 11: clarify 'our users'

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+  to have immediate access to the released changes.
+  In many scenarios it might not be the case.
+  Furthermore, creating release candidate version instead of official release
+  gives us more time to "sit" on the introduced changes, which in turn
+  increases a chance for bug detection and gives room to introduce breaking
+  changes which might be necessary to address the uncovered issue.
+2. Create official release version (restricted access). \
```
pawel-nobl9, line 18: clarify restricted access, you can be transparent that this path is only for Nobl9ers

pawel-nobl9: maybe the whole document should be divided in two sections, the first meant to open source contributors, the second for Nobl9ers

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+  Reach out to Foundations domain if you want to go with the official release.
+
+## Merging to main
+
+**NEVER** merge commits or tags based versions of n9 if they are based off of
+any other branch then main.
```
pawel-nobl9, line 24: then -> than

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+- New endpoint which is already available through n9 API.
+- Dependency update.
+- Documentation enhancement.
+
+### The changes are required to be in sync with n9 (backwards compatible)
+
+This scenario can be divided into two cases, depending on the following
+characteristics of the code you've written:
+
+- Visibility, is it exported (available through the public contract of the SDK)
```
pawel-nobl9, line 55: not sure if 'visibility' is a good word here, it's more about 'exposure' or 'accessibility' or just 'Public / Private', 
the most common wording imho is 'public / private access modifiers' but on the other hand accessibility is a term that is used for assisted technologies and things like WCAG standard

pawel-nobl9: isn't this just about changing the current public contract?

pawel-nobl9 (changes requested):

docs/RELEASE.md
nieomylnieja: I'm not sure we need to mention open source contributors in this README at all right now we don't even have contribution guidelines yet

docs/RELEASE.md
nieomylnieja: I think list is better suited here, if there would be longer descriptions and other nested lists inside I'd use new heading

daniel-zelazny (approved):
great job, well described examples :rocket:

docs/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+  Furthermore, creating release candidate version instead of official release
+  gives us more time to "sit" on the introduced changes, which in turn
+  increases a chance for bug detection and gives room to introduce breaking
+  changes which might be necessary to address the uncovered issue.
+2. **Create official release version (restricted access).** \
+  Reach out to Foundations domain if you want to go with the official release.
+
+## Merging to main
+
+**NEVER** merge commits or tags based versions of n9 if they are based off of
```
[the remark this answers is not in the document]
nieomylnieja: https://www.merriam-webster.com/grammar/based-on-vs-based-off

pawel-nobl9 (approved):

PR #314, opened 2024-03-07
feat: PC-11485 Expose Jitter, Timeout and Interval in Direct/Agent 

dawidwisn:
Expose Jitter, Timeout and Interval in Direct/Agent  v1alpha models (read only)

## Motivation

We would like to expose the values of parameters: Jitter, Timeout and Interval, which are set in Direct/Agent

## Summary

Extend Direct/Agent models

## Release Notes

New fields in Agent/Direct struct.
`Interval` The query interval sets how often the Agent queries a data source.
`Timeout` Timeout refers to the duration the Agent will wait for a response after sending a request to a data source.
`Jitter` Jitter defines a maximum random delay for each call to avoid spikes to a data source.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,9 @@ go 1.21
 
 require (
 	github.com/BurntSushi/toml v1.3.2
-	github.com/MicahParks/jwkset v0.5.14
-	github.com/MicahParks/keyfunc/v3 v3.2.8
-	github.com/aws/aws-sdk-go v1.50.35
+	github.com/MicahParks/jwkset v0.5.15
+	github.com/MicahParks/keyfunc/v3 v3.2.9
+	github.com/aws/aws-sdk-go v1.50.36
 	github.com/bmatcuk/doublestar/v4 v4.6.1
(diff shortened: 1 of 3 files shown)


manifest/v1alpha/agent/agent.go
```diff
@@ -69,6 +69,9 @@ type Spec struct {
 	Honeycomb               *HoneycombConfig                 `json:"honeycomb,omitempty"`
 	HistoricalDataRetrieval *v1alpha.HistoricalDataRetrieval `json:"historicalDataRetrieval,omitempty"`
 	QueryDelay              *v1alpha.QueryDelay              `json:"queryDelay,omitempty"`
+	Interval                *v1alpha.Interval                `json:"interval,omitempty"`
+	Timeout                 *v1alpha.Timeout                 `json:"timeout,omitempty"`
+	Jitter                  *v1alpha.Jitter                  `json:"jitter,omitempty"`
```
marcinlawnik, line 74: I would add a comment that they're read-only for now.

marcinlawnik (approved):

PR #315, opened 2024-03-07
fix: PC-12276 Prevent alert policies with alertingWindow 0m

piotrkwarcinski:
## Motivation

Prevent an edge case where users are allowed to define `alertingWindow` duration equal to `0m` as the minimal duration is defined as `5m`.

## Summary

* Added validation for the edge case.

## Testing

- Added unit tests


nobl9-adam-szymanski (approved):
LGTM

nieomylnieja (approved):

PR #330, opened 2024-03-18
feat: PC-11745 Add Primary to Objective

natalialanga:
## Motivation

Expose new field `primary` to highlight a specific objective in an SLO.


lukasz-dobek (commented):
Shouldn't we extend `manifest/v1alpha/slo/validation.go` with validation of primary objective? For example: only up to one objective can be set as primary.

lukasz-dobek (approved):
LGTM 🚀

PR #342, opened 2024-03-26
feat: PC-12282 Introduce metadata annotations

skrolikiewicz:
## Release Notes
We added the ability to define annotations on `SLOs`, `Services`, `Projects`, and `Alert Policies`.
Similarly to Kubernetes, users can now use either labels or annotations to attach metadata to objects.

The changes are compatible with the Nobl9 `1.75`, which will be released on `11.04.2024`.

## Summary
This PR introduces a mechanism for users to define annotations on SLOs, Services, Projects, and Alert Policies (the same objects to which we can assign labels).
Similarly to Kubernetes, users can use either labels or annotations to attach metadata to objects.

Annotations, similarly to k8s annotations, are defined as key/value maps:

```json
"metadata": {
  "annotations": {
    "key1" : "value1",
    "key2" : "value2"
  }
}
```


manifest/v1alpha/metadata_annotations.go
```diff
@@ -0,0 +1,68 @@
...
+	maxAnnotationKeyLength   = 63
+	minAnnotationValueLength = 1
+	maxAnnotationValueLength = 1050
+)
+
+var (
+	annotationKeyRegexp = regexp.MustCompile(`^\p{L}([_\-0-9\p{L}]*[0-9\p{L}])?$`)
+)
+
+func (a MetadataAnnotations) validateKey(key annotationKey) error {
```
nieomylnieja, line 47: I know we don't have map iterator support (yet) in `validation` pkg, but you can still define validation for each key dynamically and treat it as a separate object, that's how the map iterator would behave anyway, each key would be a new object as for JSON/YAML path there's no difference how you define that in code

nieomylnieja (changes requested):

PR #343, opened 2024-03-26
feat: add GetUser() to return current user

shubhindia:
## Motivation

We are writing an internal API where we need to retrieve current user for nobl9. Now, since it is not exposed outside the module, there was no better way to do this. Someone else might need this as well.

## Summary

adds a `GetUser` func which returns user

## Release Notes

Added `GetUser` method for `Client` which returns the current session user read from JWT token.

### DIFF
--- a/sdk/credentials.go
+++ b/sdk/credentials.go
@@ -100,7 +100,7 @@ func (c *credentials) GetOrganization(ctx context.Context) (string, error) {
 
 // GetUser first ensures a token has been parsed before returning the user,
 // as it is extracted from the token claims.
-// credentials.m2mProfile.User should not be accessed directly, but rather through this method.
+// credentials.<profile>.User should not be accessed directly, but rather through this method.
 func (c *credentials) GetUser(ctx context.Context) (string, error) {
 	if _, err := c.refreshAccessToken(ctx); err != nil {
 		return "", errors.Wrap(err, "failed to get user")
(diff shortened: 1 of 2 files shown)


sdk/credentials.go
```diff
@@ -98,6 +98,22 @@ func (c *credentials) GetOrganization(ctx context.Context) (string, error) {
 	return c.organization, nil
 }
 
+// GetUser first ensures a token has been parsed before returning the user,
+// as it is extracted from the token claims.
+// credentials.m2mProfile.User should not be accessed directly, but rather through this method.
```
nieomylnieja, line 103: To make it more generic, as we either access m2m or agent profiles:

```suggestion
// credentials.<profile>.User should not be accessed directly, but rather through this method.
```

nieomylnieja (changes requested):
Thanks a lot for the contribution! (the first external one too! :1st_place_medal:)

Please include the following note under `## Release Notes` header in the PR body:

```md
## Release Notes

Added `GetUser` method for `Client` which returns the current session user read from JWT token.
```

nieomylnieja:
The pull request template must also adhere to the rules specified in https://github.com/nobl9/nobl9-go/blob/main/.github/workflows/pr-title.yml.

So the correct title would be `feat: add GetUser() to return current user` ;)

nieomylnieja (approved):
Looks good :) Thanks again for this contribution!

nieomylnieja:
@shubhindia New version containing your changes has been released and ready to use: https://github.com/nobl9/nobl9-go/releases/tag/v0.79.0

PR #345, opened 2024-03-27
feat: Add rules for maps in validation

nieomylnieja:
## Motivation

Currently we're lacking any constructs in `validation` pkg to verify maps. While it can be done more "manually" it would be nice to have a similar validator to current `ForEach`, which should probably be renamed to `ForSlice`.


skrolikiewicz (approved):
Awesome :D I'm gonna use it right away!

PR #347, opened 2024-03-28
fix: Decode to json.Number instead of float

nieomylnieja:
## Motivation

When working with generic object representation, like `v1alpha.GenericObject` it was impossible to properly handle the data retaining integer and float distinction as Golang's `encoding/json` always converts JSON numbers to floats.

The solution is retain the original number format through `json.Number` and user custom YAML encoding for the `json.Number` type.

## Release Notes

When reading JSON encoded `manifest.Object`, the numbers are now retained in `json.Number` format instead of being converted to floats.
Added `PrintObjects` and `PrintObject` which handle `json.Number` correctly when encoding to YAML format.

### DIFF
--- a/internal/examples/client.go
+++ b/internal/examples/client.go
@@ -1,13 +1,12 @@
 package examples
 
 import (
-	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"os"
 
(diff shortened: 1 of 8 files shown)


dawidwisn (approved):
lgtm

PR #355, opened 2024-04-04
feat: Allow for mocking sdk.Client

nieomylnieja:
## Motivation

Currently it's impossible to mock `sdk.Client` due to how versioned endpoints groups are returned, for example: `sdk.Client.Objects` returns a concrete struct: `Versions` which in turn holds methods for extracting specific API version's endpoints. There's no point mocking the `sdk.Client.Objects` method, it doesn't give us anything as it's just an intermediary on our way to access the specific endpoint function.

## Summary

Changed the `endpoints` package to return interfaces rather than concrete structs so that they can be mocked.

Used https://github.com/vburenin/ifacemaker for the interfaces generation as it lifts up the maintenance cost of adding/modifying API methods.


daniel-zelazny (approved):
Good job - but we have some vulnerabilities reported.

PR #359, opened 2024-04-05
feat: Improve labels validation

nieomylnieja:
## Summary

Improved `v1alpha.Labels` validation which now relies entirely on `validation` package ensuring consistent and informative errors.

Removed old checks for whole error message in object's validation in favour of regex check.

Labels validation is now performed against each object supporting labels.


internal/manifest/v1alphatest/doc.go
skrolikiewicz, line 1: cool :)

skrolikiewicz (approved):
Let's goo

PR #369, opened 2024-04-12
feat: Stabilize validation API - When usage

nieomylnieja:
## Motivation

In order to move validation package to a separate library we need to first stabilize it's API.

This is the first PR in a cycle of changes required to achieve that.


daniel-zelazny (approved):
Looks good to me :+1:

PR #373, opened 2024-04-17
chore: Stabilize validation API - remove StopOnError and simplify internal logic

nieomylnieja:
## Motivation

In order for validation pkg to mature into it's own library we have to cleanup the current hacks and non-ideal solutions.

## Summary

- Removed hacky `StopOnError` construct
- Simplified maps and slices validators' logic
- Added `CascadeMode` similar to https://docs.fluentvalidation.net/en/latest/cascade.html

### DIFF
--- a/manifest/v1alpha/direct/validation.go
+++ b/manifest/v1alpha/direct/validation.go
@@ -24,7 +24,7 @@ var directValidation = validation.New[Direct](
 var specValidation = validation.New[Spec](
 	validation.For(validation.GetSelf[Spec]()).
 		Rules(exactlyOneDataSourceTypeValidationRule).
-		StopOnError().
+		CascadeMode(validation.CascadeModeStop).
 		Rules(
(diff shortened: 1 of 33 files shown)


manifest/v1alpha/direct/validation.go
```diff
@@ -24,7 +24,7 @@ var directValidation = validation.New[Direct](
 var specValidation = validation.New[Spec](
 	validation.For(validation.GetSelf[Spec]()).
 		Rules(exactlyOneDataSourceTypeValidationRule).
-		StopOnError().
+		Cascade(validation.CascadeModeStop).
```
skrolikiewicz, line 27: Shouldn't that be placed before `Rules(exactlyOneDataSourceTypeValidationRule).`

nieomylnieja: yes :) thanks

skrolikiewicz (approved):
👍

PR #375, opened 2024-04-18
feat: Add validation plan generator

nieomylnieja:
## Motivation

When documenting an API it's often a struggle to keep consistency between the code and documentation we write for it.
What If your code could be self descriptive? Specifically, what If we could generate documentation out of our validation rules?

This PR explores this topic attempting to create a builtin mechanism for generating a validation plan, which could then be used to generate human-readable documentation for validation rules.

You can generate a validation plan by running:

```sh
go run internal/cmd/docgen/generator.go
```

This PR **does not** attempt to make the whole validation plan fully polished and ready to go, but rather create a foundation on top of which to build on at the same time proving it can be done.


skrolikiewicz (approved):
Looks like a solid foundation :)

PR #377, opened 2024-04-19
PC-12652 feat: Allow to set any mathematical operator for burnedBudget alerting condition

nobl9-adam-szymanski:
## Motivation

Allow to provide any mathematical operator for `burnedBudget` alerting measurement ("Remaining error budget is ... ") or allow to leave it empty. If no operator is provided (empty field), use default operator correlated with given measurement.

This can be used with combination with other condition types to be alerted when there is error budget or there is no more error budget left to burned.

## Summary

- For condition type "Remaining error budget" any operator type is allowed.
- For condition types  "Time to burn budget", "Time to burn entire budget", "Average burn rate" there is one type of operator allowed (or default).

## Testing

1. Confirm that validation passes custom mathematical operator.
2. Confirm that it's saved into database - you should see provided operator when for example using sloctl `sloctl get alertpolicy <name>`.

User wants to be alerted when there is no budget left with `timeToBurnEntireBudget` measurement ("Entire error budget would be exhausted in ... "). 

When there is still a budget to burn user prefers to be alerted with `timeToBurnBudget` measurement ("Remaining error budget would be exhausted in ... ").

```yaml
---
# There is no budget left and entire budget would be exhausted in 8h and this condition lasts for 15m.
apiVersion: n9/v1alpha
kind: AlertPolicy
metadata:
  name: entire-exhaustion-prediction-8h
  displayName: Entire budget exhaustion in 8h
  labels:
    type:
      - time-exhaustion
spec:
  description: Entire error budget allocation prediction for 99%, 30 Day Rolling.
  severity: Low
  coolDown: "15m"
  conditions:
    - measurement: timeToBurnEntireBudget
      value: "8h"
      lastsFor: "15m"
    - measurement: burnedBudget
      value: 1
      op: gte
---
# There is a budget to burn, remaining budget would be exhausted in 3d and this condition lasts for 15m.
apiVersion: n9/v1alpha
kind: AlertPolicy
metadata:
  name: remaining-exhaustion-prediction-3d
  displayName: Remaining budget exhaustion in 3d
  labels:
    type:
      - time-exhaustion
spec:
  description: Remaining error budget Allocation prediction
  severity: Medium
  coolDown: "15m"
  conditions:
    - measurement: timeToBurnBudget
      value: "72h"
      lastsFor: "15m"
    - measurement: burnedBudget
      value: 1
      op: lt
```

## Release Notes

User may now set any valid mathematical operator (`lte`, `lt`, `gte`, `gte`) for measurement `burnedBudget`. This can be useful to combine it with other measurements like `timeToBurnBudget` or `averageBurnRate` to be alerted only when there is error budget left or only when whole error budget is already burned.


manifest/v1alpha/alertpolicy/measurement.go
```diff
@@ -63,6 +63,16 @@ func GetExpectedOperatorForMeasurement(measurement Measurement) (v1alpha.Operato
 	}
 }
 
+// getExpectedOperatorForMeasurement returns the operator that should be paired with a given measurement.
+func getExpectedOperatorForMeasurement(measurement Measurement) (v1alpha.Operator, error) {
```
nobl9-adam-szymanski, line 67: Actually, we can leave it public to avoid breaking change if anyone except us use it.

nobl9-adam-szymanski, line 67: I don't think that anyone except Nobl9 use this method to get default operator which is optional, but we could leave it as public. I think the risk that anyone use it is very low and we can try to change it to private and wait for possible feedback.

PR #412, opened 2024-05-15
chore: Add option to change output format to compacted JSON in docgen

nieomylnieja:
## Motivation

Currently `docgen` only outputs yaml, which is slower to decode and weights more but it's easier to read by a human.
External tools however, would benefit more from having it in JSON than YAML.


daniel-zelazny (approved):
LGTM :)

PR #422, opened 2024-05-23
feat: add createdAt and createdBy to project

shubhindia:
## Motivation

Right now there is no info for the project about when it was created and who created it. We are developing some internal APIs to manage nobl9, and would love to get this info added. 

## Summary
I have added relevant fields to the `Project` struct, the core controller needs to populate these fields, but since it isn't open-source, I can't do that myself.
@nieomylnieja  Let me know your thoughts 😄 

## Release Notes

Added `createdAt` and `createdBy` fields to v1alpha Project.
These fields are not yet populated by the API, follow our [release notes](https://docs.nobl9.com/application-release-notes) to get informed when the API support ships.

### DIFF
--- a/manifest/v1alpha/project/project.go
+++ b/manifest/v1alpha/project/project.go
@@ -44,6 +44,8 @@ type Metadata struct {
 
 // Spec holds detailed specification of the Project.
 type Spec struct {
+	CreatedAt string `json:"createdAt,omitempty"`
+	CreatedBy string `json:"createdBy,omitempty"`
 	// Description allows for a more detailed description of the Project.
 	Description string `json:"description" validate:"description" example:"Bleeding edge web app"`
 }


nieomylnieja:
hey @shubhindia ;)

This is certainly doable, we need to consult this idea in our team responsible for object's API though :)
If we're to add that to Project, we should include that for all of our objects too.

I'll let you know once we have some thoughts/details on this one, thanks for bringing this up!

shubhindia:
> hey @shubhindia ;)
> 
> This is certainly doable, we need to consult this idea in our team responsible for object's API though :) If we're to add that to Project, we should include that for all of our objects too, especially since SLO already has it.
> 
> I'll let you know once we have some thoughts/details on this one, thanks for bringing this up!

Yup. Ideally it should be available for all the objects, but since I qas working on projects, I added for them. Let me know if you guys decide to add it for other objects too. Happy to make that change here 😄

nieomylnieja:
We talked through this today and we'd want to support those fields on most of our objects.
Since you're only interested in `Project` right now, I'd limit this PR scope to it alone.

I'll work on the changes for our internal system to support returning these fields, once that's done we'll merge this one.

Afterwards I'll work on extending support for these fields for other objects as well :)

shubhindia:
> We talked through this today and we'd want to support those fields on most of our objects. Since you're only interested in `Project` right now, I'd limit this PR scope to it alone.
> 
> I'll work on the changes for our internal system to support returning these fields, once that's done we'll merge this one.
> 
> Afterwards I'll work on extending support for these fields for other objects as well :)

Cool. Sounds good 😁

nieomylnieja:
The changes are available in https://github.com/nobl9/nobl9-go/releases/tag/v0.82.0.

PR #476, opened 2024-07-01
chore: Correct Alert Method type names

nieomylnieja:
## Motivation

Having sth like `v1alpha.Email` in the namespace is way too vague, we need more specific naming, that's why we need a prefix.


daniel-zelazny (approved):
LGTM

BSski (approved):

PR #477, opened 2024-07-01
feat: Define manifest objects examples in code

nieomylnieja:
## Motivation

We're currently lacking in examples for our `v1alpha` objects.
This PR covers that hole by adding utilities for generating these examples and verifying them.

## Release Notes

Added extended examples to the `v1alpha` objects with different configuration variants.


daniel-zelazny (approved):
nice work :rocket:

PR #482, opened 2024-07-05
chore: Switch go-yaml library to Nobl9 fork

nieomylnieja:
## Motivation

Modules using `nobl9-go` cannot rely on `replace` clause in the SDK if they are using `go-yaml` package as well, in such scenario these modules have to using `replace` clause too and this in turn breaks `go install`, here's an example:

```
go install github.com/nobl9/sloctl/cmd/sloctl@latest
go: downloading github.com/nobl9/sloctl v0.3.2
go: github.com/nobl9/sloctl/cmd/sloctl@latest (in github.com/nobl9/sloctl@v0.3.2):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-un:
+run:
   timeout: 5m
   modules-download-mode: readonly
   skip-dirs-use-default: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,8 +1,6 @@
-run:
+un:
   timeout: 5m
   modules-download-mode: readonly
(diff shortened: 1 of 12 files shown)


.golangci.yml
```diff
@@ -1,8 +1,6 @@
-run:
+un:
```
kskitek, line 1:
```suggestion
run:
```

kskitek (approved):

PR #490, opened 2024-07-10
chore: Add v1alpha Agent and Direct end-to-end tests

nieomylnieja:
Adds v1alpha `Agent` and `Direct` tests.

### DIFF
--- a/tests/v1alpha_project_test.go
+++ b/tests/v1alpha_project_test.go
@@ -104,7 +104,6 @@ func generateV1alphaProject(t *testing.T) v1alphaProject.Project {
 
 func assertV1alphaProjectsAreEqual(t *testing.T, expected, actual v1alphaProject.Project) {
 	t.Helper()
-	expected = deepCopyObject(t, expected)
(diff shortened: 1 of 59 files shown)


tests/v1alpha_project_test.go
```diff
@@ -104,7 +104,6 @@ func generateV1alphaProject(t *testing.T) v1alphaProject.Project {
 
 func assertV1alphaProjectsAreEqual(t *testing.T, expected, actual v1alphaProject.Project) {
 	t.Helper()
-	expected = deepCopyObject(t, expected)
```
daniel-zelazny, line 107: could we get rid of it in other places?

daniel-zelazny (approved):
Looks good, consider one remark.

tests/v1alpha_project_test.go
nieomylnieja: I got rid of it there cause expected object was not modified, we need it in other places due to data races, these tests are run in parallel but they operate on a single object instance

PR #494, opened 2024-07-17
chore: Add SLO end-to-end tests

nieomylnieja:
Added SLO end-to-end tests.

The bigger object batches are run in batches, unfortunately for some of the objects it is not possible to run multiple `apply` operations concurrently, so we need to do it synchronously.

The tested SLOs operate on static Agent and Direct sets which are applied before each test but do are not removed after the tests finish, they also have static names and project, so the apply will not create any new objects if they were already applied.

## Extra changes

- switched SLO examples to use Agent instead of Direct as this is safer, since not all integrations support Direct mode
- minor issues with Agent and Direct examples
- fixed `docgen` which was not working with the current examples
- fixed bad over total metric examples which were generated incorrectly
- added new line in `http.go` for retry log which was there once before but got removed by accident


daniel-zelazny (approved):
Awesome :medal_military:

PR #516, opened 2024-08-12
chore: PC-13744 Composite SLOs in examples and SDK e2e tests

ditrytus:
## Motivation

Recently added SDK e2e tests were developed when Composite SLOs were not released yet. We need to close that test coverage gap.

## Summary

The code that generates examples of SLOs now includes all variants for Composite SLOs. Few cases required special handling:
* Composite SLOs don't have `indicator` section and don't have an association with data sources
* Composite SLOs depend on other SLOs. Normal SLOs depend only on objects of other kinds.

YAML examples were regenerated from updated code.

`Makefile` contains small improvement that when running e2e tests locally `NOBL9_SDK_TEST_RUN_SEQUENTIAL_APPLY_AND_DELETE` env var is set to `true`, because usually, tests run locally are run against smaller and slower dev environments.

Fixed `check/generate` target in `Makefile` that was not checking generated yaml examples.

## Testing

Locally:
```
make test/e2e
```

Run "End-to-end tests dispatch" workflow with and without sequential requests option:

https://github.com/nobl9/nobl9-go/actions/runs/10355310664/job/28662650970
https://github.com/nobl9/nobl9-go/actions/runs/10355329422/job/28662712884


internal/manifest/v1alpha/examples/slo_composite.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (s sloCompositeExample) String() string {
+	return fmt.Sprintf(
+		"Composite SLO using %s budgeting method and %s time window",
+		s.BudgetingMethod,
+		s.TimeWindowType,
+	)
+}
+
+func (s sloCompositeExample) Slo() v1alphaSLO.SLO {
```
nieomylnieja, line 43: Let's stick to the conventional SLO naming, i.e. all capital letters:

```suggestion
func (s sloCompositeExample) SLO() v1alphaSLO.SLO {
```

internal/manifest/v1alpha/examples/slo_composite.go
```diff
@@ -0,0 +1,103 @@
...
+									Project:     "project-2",
+									SLO:         "slo-2",
+									Objective:   "objective-2",
+									Weight:      2,
+									WhenDelayed: v1alphaSLO.WhenDelayedCountAsBad,
+								},
+								{
+									Project:     "project-3",
+									SLO:         "slo-3",
+									Objective:   "objective-3",
```
nieomylnieja, line 92: can we come up with some nicer, down to earth names than that?
these examples end up in our docs, It'd be nice to try and make them more real-world than "examply"

internal/manifest/v1alpha/examples/labels.go
```diff
@@ -16,3 +16,12 @@ func exampleLabels() v1alpha.Labels {
...
 }
+
+func exampleCompositeLabels() v1alpha.Labels {
+	return v1alpha.Labels{
+		"team":   {"green", "ux"},
+		"env":    {"prod", "dev"},
+		"region": {"us", "eu"},
+		"area":   {"user-experience"},
+	}
+}
```
nieomylnieja, line 27: since this is only used in one place by composite SLO, it doesn't need a separate function, you can just inline it in the generation code, or move the function to `slo_composite.go` file

nieomylnieja (changes requested):

nieomylnieja (approved):

PR #520, opened 2024-08-13
fix: PC-13744 Removed anomaly config from composite e2e test

ditrytus:
## Summary

Removed anomaly config from composite tests


piotrkwarcinski (changes requested):
looks bad

n-lechowicz (approved):
🎁

PR #531, opened 2024-08-30
feat: Expose structured API response errors

nieomylnieja:
## Motivation

`sdk.Client` methods which interact with Nobl9 API currently return only generic text based errors.
This obstructs the relevant details from the user. Furthermore, the current default text format for these errors is fairly poor and doesn't contain details like endpoint path.

## Summary

- Added `sdk.APIError` which is a structured error that contains all the relevant API error details as well as produces an improved (over the previous state) text representation.
- Improved docs.
- Removed PlantUML in favour of Mermaid, since it's much easier to use as it doesn't require Java and renders natively in Markdown.

## Testing

Covered by both unit and end-to-end tests.

## Release Notes

`sdk.Client` API calls now can return `sdk.APIError`, which is a structured error providing details of the error returned by Nobl9's API.


sdk/client_errors.go
```diff
@@ -0,0 +1,80 @@
...
+	Message    string `json:"message"`
+	StatusCode int    `json:"statusCode"`
+	Method     string `json:"method"`
+	URL        string `json:"url"`
+	TraceID    string `json:"traceId,omitempty"`
+}
+
+// IsConcurrencyIssue returns true if the underlying API error is a concurrency issue.
+// If so, the operation can be retried.
+func (r APIError) IsConcurrencyIssue() bool {
```
skrolikiewicz, line 26: I imagine there might be other kinds of errors that are not entirely related to concurrent issues.
As a consumer, I'm more interested in whether a request that ended up with an error can be retried rather than whether it was caused by concurrent requests (which honestly should be 409).

Maybe we could introduce a property similar to the one you can find in aws sdk (`IsErrorRetryable` - https://github.com/aws/aws-sdk-go-v2/blob/main/aws/retry/retryable_error.go#L15)

nieomylnieja: that's a great thought! I agree :)

skrolikiewicz (approved):
🚀

PR #551, opened 2024-09-23
test: PC-13893 Make E2E tests forward compatible

ditrytus:
## Motivation

`value` field of `objective` for Composite SLO is being [deprecated](https://github.com/nobl9/nobl9-go/pull/549). Before the update in SDK can be made, a N9 platform first needs to be [updated](https://github.com/nobl9/n9/pull/15406). To be able to release changes in the platform, e2e tests of the latest released SDK need to pass against it.

## Summary

E2E test assertions will pass with and without [changes](https://github.com/nobl9/n9/pull/15406) in N9 platform.

## Related changes

https://github.com/nobl9/nobl9-go/pull/551
https://github.com/nobl9/nobl9-go/pull/549
https://github.com/nobl9/n9/pull/15406
https://github.com/nobl9/terraform-provider-nobl9/pull/312
https://github.com/nobl9/terraform-provider-nobl9/pull/295

## Testing

Run `make test/e2e` on both environments containing and not containing https://github.com/nobl9/n9/pull/15406.

### DIFF
--- a/tests/v1alpha_slo_test.go
+++ b/tests/v1alpha_slo_test.go
@@ -293,5 +293,8 @@ func assertV1alphaSLOsAreEqual(t *testing.T, expected, actual v1alphaSLO.SLO) {
 	actual.Spec.CreatedBy = ""
 	actual.Status = nil
 	actual.Spec.TimeWindows[0].Period = nil
+	if actual.Spec.Objectives[0].Value == nil {
+		actual.Spec.Objectives[0].Value = ptr(0.0)
+	}
 	assert.Equal(t, expected, actual)
 }


tests/v1alpha_slo_test.go
```diff
@@ -293,5 +293,8 @@ func assertV1alphaSLOsAreEqual(t *testing.T, expected, actual v1alphaSLO.SLO) {
 	actual.Spec.CreatedBy = ""
 	actual.Status = nil
 	actual.Spec.TimeWindows[0].Period = nil
+	if actual.Spec.Objectives[0].Value == nil {
+		actual.Spec.Objectives[0].Value = ptr(0.0)
+	}
```
piotrkwarcinski, line 298: Shouldn't this be only set when the SLO is composite SLO?

piotrkwarcinski: or value will be always only `nil` if the SLO is a composite SLO?

ditrytus: Could: yes, should: maybe, must: no. This change needs to live only for a few days, so I just didn't bother. But... you're the reviewer here, so I'll leave it to your decision.

ditrytus: Before changes to N9, `value` is never `nil`. After changes to N9, `value` can only be `nil` for Composite SLOs which will be tested in https://github.com/nobl9/nobl9-go/pull/549. Value can be not `nil` if Composite SLO was created prior to change and had non-zero value.

piotrkwarcinski: > After changes to N9, value can only be nil for Composite SLOs
that's good enough for me :)

piotrkwarcinski (approved):

PR #552, opened 2024-09-23
fix: validate Splunk Observability release channel

mkaras-nobl9:
## Motivation

Allow to create Splunk Observability Directs in release channel alpha.

## Summary

Adds custom validation for the Direct `releaseChannel` property when using Splunk Observability.

## Testing

- The validation should allow you to create Splunk Observability Direct with the `alpha` value in the release channel prop. Any other value should fail.
- Only Splunk Observability Directs can use the `alpha` release channel.


manifest/v1alpha/direct/validation.go
```diff
@@ -223,6 +224,7 @@ var (
 const (
 	errCodeExactlyOneDataSourceType              = "exactly_one_data_source_type"
 	errCodeQueryDelayGreaterThanOrEqualToDefault = "query_delay_greater_than_or_equal_to_default"
+	errUnsupportedReleaseChannel                 = "unsupported_release_channel"
```
nieomylnieja, line 227:
```suggestion
	errCodeUnsupportedReleaseChannel                 = "unsupported_release_channel"
```

manifest/v1alpha/direct/validation.go
```diff
@@ -383,6 +385,29 @@ var queryDelayGreaterThanOrEqualToDefaultValidationRule = govy.NewRule(func(spec
 	return nil
 }).WithErrorCode(errCodeQueryDelayGreaterThanOrEqualToDefault)
 
+const errorCodeUnsupportedReleaseChannel = "unsupported_release_channel"
```
nieomylnieja, line 388: duplicated entry, see above

manifest/v1alpha/direct/validation.go
```diff
@@ -383,6 +385,29 @@ var queryDelayGreaterThanOrEqualToDefaultValidationRule = govy.NewRule(func(spec
...
+			spec.ReleaseChannel,
+			errors.New("must be one of [stable, beta]"),
+		)
+	}
+
+	if typ == v1alpha.SplunkObservability && spec.ReleaseChannel != v1alpha.ReleaseChannelAlpha {
+		return govy.NewPropertyError(
+			"releaseChannel",
+			spec.ReleaseChannel,
+			errors.New("must be alpha for Splunk Observability"),
```
nieomylnieja, line 404: just to highlight the expected value better:

```suggestion
			errors.New("must be 'alpha' for Splunk Observability"),
```

manifest/v1alpha/direct/validation.go
```diff
@@ -383,6 +385,29 @@ var queryDelayGreaterThanOrEqualToDefaultValidationRule = govy.NewRule(func(spec
 	return nil
 }).WithErrorCode(errCodeQueryDelayGreaterThanOrEqualToDefault)
 
+const errorCodeUnsupportedReleaseChannel = "unsupported_release_channel"
+
+var releaseChannelValidationRule = govy.NewRule(func(spec Spec) error {
+	typ, _ := spec.GetType()
+	if typ != v1alpha.SplunkObservability && spec.ReleaseChannel == v1alpha.ReleaseChannelAlpha {
```
nieomylnieja, line 392: are we sure about this? I know `alpha` is reserved to be set by customer support on demand, but even so, such Direct will not be allowed an update if we do this. It seems you just added this rule I don't see it being present before

nieomylnieja (changes requested):

manifest/v1alpha/direct/validation.go
mkaras-nobl9: Yes, I added this on purpose to preserve the alpha channel for SplunkObs since no one is using it. We want to separate the Splunk Observability workload and prevent other data sources from running on the same DE instance. And since user with [this change](https://github.com/nobl9/nobl9-go/pull/552/files#diff-12637746a46ee5307e59e4b5ebbd6e7121055456c60bb575d8a34488f5e99369L37) can set alpha channel this condition is required. This is not a permanent change, but we need here for 2-3 months or so to implement better mechanism to run data sources on dedicated instances. When it comes to usability, currently user also can't modify it's direct when it's in alpha channel, this change will at least allow to create and modify SplunkObs in alpha channel.

nieomylnieja (approved):

PR #560, opened 2024-09-27
fix: add max query delay validation per data source

mkaras-nobl9:
## Motivation

Splunk Observability max query delay is <= 15 minutes, but it used default 24h for all data sources.

## Summary

- validate Splunk Observability query delay is <= 15m
- any other not defined data source uses default 24 hours query delay as max value
- sort manifest/v1alpha/data_sources.go by functionality

## Release Notes

Adds max query delay validation for Splunk Observability  to 15 minutes.


bdw-nobl9 (approved):

paulina-n9:
lgtm

PR #585, opened 2024-11-05
feat: Add Replay support to Elasticsearch

nuusk:
## Motivation

Adding Replay support to Elasticsearch integration.

## Summary

Adds data retrieval definitions for Elasticsearch using Agent Connection.
Since Elasticsearch cannot be conffigured via Direct connection, those definitions for direct were not changed.

## Release Notes

Adds Replay support to Elasticsearch data source (Agent connection only).


dawidwisn (approved):
lgym

PR #591, opened 2024-11-19
feat: Update Honeycomb config to support only single query good over total [PC-14879]

nieomylnieja:
## Release Notes

Added `spec.objectives[*].goodTotal.honeycomb` single query count metrics support for Honeycomb SLOs.

## Breaking Changes

Honeycomb SLO now **ONLY** supports `spec.objectives[*].goodTotal.honeycomb` single query count metrics.
Furthermore, the `calculation` property of Honeycomb metric is no longer supported.


nikodemrafalski (approved):

nikodemrafalski:
Since the "old" honeycomb contact is still allowed, could you add at least one ratio example using calculation field, same goes for tests

nikodemrafalski (approved):
reviewed: https://github.com/nobl9/nobl9-go/pull/591/commits/6e4462a33f65776fd25aa13ff526764158e5a150

PR #613, opened 2024-12-15
fix: Adjust validation for missing metric spec [PC-15176]

nieomylnieja:
## Release Notes

When an SLO is defined with missing metric spec for either count or raw metrics, the SDK validation will now return an error.


skrolikiewicz (approved):
👍

PR #615, opened 2024-12-24
feat: PC-15221 Add EndTime to replay

marcinlawnik:
## Motivation

With replay queues we now save end time to be more precise than Duration (as a replay can wait in queue before it is started).

## Summary

Changed ReplayStatus and ProcessStatus structs

## Testing

Test in sloctl and in UI.

## Release Notes

End time is now returned for new replays in progress and finished.


bdw-nobl9 (approved):

paulina-n9:
Tested both in sloctl and in UI and LGTM

PR #617, opened 2025-01-07
chore: Improve sdk.ReadObjects docs [PC-15261]

nieomylnieja:
## Motivation

It can be unclear how the `sdk.ObjectSourceTypeURL` is interpreted, we should make the constraints it operates within clear.

## Summary

Improved linking of the Go docs.


daniel-zelazny (approved):
:+1:

PR #620, opened 2025-01-09
feat: Add filtering by service name to GetV1alphaSLOs

nieomylnieja:
## Release Notes

Added option to filter SLOs by service name. The new filter is available via `v1.GetSLOsRequest.Service` field.


daniel-zelazny (approved):
:+1:

PR #633, opened 2025-01-23
feat: Update Honeycomb config to forbid raw metrics; support good total single query only [PC-15396] 

nikodemrafalski:
## Motivation

Changed the way we support Honeycomb integration - now only ratio (countMetrics) SLOs are available, using only `COUNT` calculation method. Raw (threshold) metrics are forbidden, which is tested by validation.

## Summary

Unsupported ways of configuring Honeycomb data sources were removed, and the validation checking metric type (only ratios are allowed) was introduced.


dawidwisn (approved):
lgtm! We are waiting those changes: https://github.com/nobl9/n9/pull/16773

nikodemrafalski (commented):
@bdw-nobl9 commits LGTM

bdw-nobl9 (approved):

PR #639, opened 2025-02-03
chore: Run cross platform unit tests

nieomylnieja:
## Motivation

In order to better support Linux, MacOS and Windows usages we should run the unit tests on each of them.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,14 +8,18 @@ on:
       - main
 jobs:
   test:
-    name: Run unit tests
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+
     steps:
       - name: Check out code
         uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - name: Install go
+        uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
           cache: false
-      - name: Run unit tests
-        run: make test
+      - name: Test
+        run: go test -race -cover ./... ./docs/mock_example


internal/stringutils/stringutils.go
```diff
@@ -0,0 +1,6 @@
+//go:build !windows
+
+package stringutils
+
+// RemoveCR by default does nothing, it only modifies the string if built for Windows.
```
daniel-zelazny, line 5:
```suggestion
// RemoveCR by default does nothing for building in  other than Windows OS.
```
I was bit confused by this comment, firstly I thought that you've missed something or the flag is wrong.

sdk/reader_test.go
```diff
@@ -447,8 +450,12 @@ func definitionsMatchExpected(t *testing.T, definitions []manifest.Object, meta
 		err := templates.ExecuteTemplate(buf, m.Name+".tpl.json", m)
 		require.NoError(t, err)
 		var decoded interface{}
-		err = json.Unmarshal(buf.Bytes(), &decoded)
-		require.NoError(t, err)
+		data := buf.Bytes()
+		if runtime.GOOS == "windows" {
```
daniel-zelazny, line 454: We handle the OS specifics using build flags in `stringutils`, I wonder if it wouldn't be better if we approach consistently and extract some OS destined files as you did with stringutils.

daniel-zelazny (approved):
Nice work, please consider two small suggestions :)

sdk/reader_test.go
nieomylnieja: note that stringutils is reusable, this is not, it's very much case specific :) furthermore it is used in tests, the  `runtime.GOOS` check is almost completely free

internal/stringutils/stringutils.go
nieomylnieja: I tried making it more explicit, please check the updated version.

PR #642, opened 2025-02-05
chore: Fix Annotations e2e tests

nieomylnieja:
## Summary

Added missing `createdBy` check to the tests. The new field was introduced in https://github.com/nobl9/nobl9-go/pull/631 but tests were not updated.


daniel-zelazny (approved):
:+1:

PR #655, opened 2025-02-27
fix: Correct labels usage in Report object

nieomylnieja:
## Motivation

Report object was redefining `v1alpha.Labels` instead of using it.

## Summary

Fixed some minor issues reported by _gopls_.

## Breaking Changes

`report.Labels`, `report.LabelKey` and `report.LabelValue` have been removed, use `v1alpha.Labels` instead.


skrolikiewicz (approved):
🏷️ 🚀

PR #666, opened 2025-03-22
feat: Relax MetadataAnnotations validation

nieomylnieja:
## Release Notes

`v1alpha.MetadataAnnotations` now allows upper case characters in keys.


skrolikiewicz (approved):
👍

PR #673, opened 2025-03-31
feat: Upgrade Go to 1.24

nieomylnieja:
## Summary

- Upgraded Go from 1.22 to 1.24
- Updated golangci linter to latest v1 version
- Updated several dependencies
- Switched back to goccy/go-yaml as several previously reported issues have been fixed. The only change I noticed is quoting of dates (now they are double quoted), most likely due to the presence of `:`, which makes sense in terms of YAML semantics.

## Release Notes

Go version has been upgraded from 1.22 to 1.24


daniel-zelazny (approved):
Nicely done :+1:

PR #674, opened 2025-04-01
fix: PC-16593 Wrongly calculated calendar-aligned time windows starting at 29th-31st day of the month 

ditrytus:
## Summary

Originally number of months between 2 dates in the same year was calculated only as a difference in the month number. 

Calculating the number of months between 2 dates needs to be consistent with normalizing dates used by `AddDate`, for example:
March 31st + 1 Month = April 31st (need normalization because April only has 30 days) = May 1st
March 31st - 1 Month = February 31st (needs normalization because Februaly only has 28 days) = March 3rd.

When calendar-aligned time window has start date between 29th-31st of the month it would incorrectly calculate the number from the start to a given date, for example:
March 2nd to March 31st = 0 Months, while correct value should be -1.
March 31st to May 1st = 2, while, correct value should be 1.

## Release Notes

Fixed an error in which specific time ranges on the timeline were inappropriately assigned to the wrong period of a calendar-aligned time window which start date was between 29th and 31st day of the month.


manifest/v1alpha/twindow/twindow_test.go
```diff
@@ -565,6 +565,46 @@ func TestGetTimePeriod(t *testing.T) {
...
+				unit:          Month,
+				count:         1,
+				startTime:     "2025-01-31 00:00:00",
+				timeZone:      "UTC",
+				timestamp:     "2025-03-02T23:59:59Z",
+				expectedBegin: "2025-01-31T00:00:00Z",
+				expectedEnd:   "2025-03-03T00:00:00Z",
+			},
+			{
+				name:          "PC-16593 March 31st + 1 Month = May 1st",
```
nieomylnieja, line 599: can we have a test case with an addition of more than 1 month? :pray:

nieomylnieja (approved):

nobl9-adam-szymanski (approved):
It takes a me a while to understand this few lines of code, especially this one `time.Date(year, month+1, 0, 0, 0, 0, 0, time.UTC)` but I had to admin that this is very clever way to find the last day of the month. Anyway, it works as expected.

manifest/v1alpha/twindow/twindow_test.go
ditrytus: Thanks for that suggestion. It turned out that there were still bugs in the logic!

manifest/v1alpha/twindow/twindow.go
```diff
@@ -112,21 +112,27 @@ type monthPeriodCalculator struct{}
...
 
 	return (timestampYear-timeWindowYear)*int(monthsInYear) + int(timestampMonth) - int(timeWindowMonth) + correction
 }
 
+func isDateNormalized(timeWindowStartDay int, date Date) bool {
+	timestampYear, timestampMonth, timestampDay := date.Date()
+	daysInPreviousMonth := daysIn(timestampMonth-1, timestampYear)
+	wasDateNormalized := timeWindowStartDay > daysInPreviousMonth && timestampDay <= timeWindowStartDay-daysInPreviousMonth
+	return wasDateNormalized
+}
```
nieomylnieja, line 134: Maybe we can simplify this:

```suggestion
func normalizeDate(timeWindowStartDay int, date Date) int {
	timestampYear, timestampMonth, timestampDay := date.Date()
	daysInPreviousMonth := daysIn(timestampMonth-1, timestampYear)
	isNormalized := timeWindowStartDay > daysInPreviousMonth && timestampDay <= timeWindowStartDay-daysInPreviousMonth
  if isNormalized {
    return -1
  }
  return 0
}
```

then in both your use cases you could just do:

```go
correction := normalizeDate(timeWindowDay, date)
```

nieomylnieja (changes requested):

manifest/v1alpha/twindow/twindow.go
ditrytus: Ok, check now.

nieomylnieja (approved):

PR #685, opened 2025-04-08
feat: Allow to set `alertAfter` for SLO no data alerts

nobl9-adam-szymanski:
## Motivation

Allow users to set NO DATA `alertAfter` time for an SLO. Users now can change default 15m period.

## Summary

```yaml
- apiVersion: n9/v1alpha
  kind: SLO
  metadata:
    name: my-slo
    project: my-project
  spec:
    alertPolicies: []
    anomalyConfig:
      noData:
        alertAfter: 30m
        alertMethods:
          - name: my-alert-method-1
            project: my-project-1
          - name: my-alert-method-1
            project: my-project-2
```

## Testing

Apply and get SLO with custom alert after with `sloctl` or Terraform.

## Release Notes

Allow users to set NO DATA `alertAfter` time for an SLO. Users now can change default 15m period when notification is sent since Nobl9 received last time point for given SLO.

## Breaking Changes

New field `alertAfter` is returned for existing SLOs with configured NO DATA with default value "15m". This might be detected as a change, because this is new field, however "15m" value is used for detection now.


manifest/v1alpha/slo/validation.go
```diff
@@ -282,6 +282,23 @@ var anomalyConfigValidation = govy.New[AnomalyConfig](
...
+						return 0, minimalNoDataAlertAfterRule.Validate(value)
+					}
+					return value, err
+				}).
+				WithName("alertAfter").
+				OmitEmpty().
+				Rules(
+					rules.DurationPrecision(time.Minute),
+					minimalNoDataAlertAfterRule,
+					rules.LTE(31*time.Hour*24)),
```
nieomylnieja, line 301: While this is a valid working approach, I think we can improve it a bit :)
I think the main culprit here is that `AlertAfter` field should really be a pointer, If we make that change it will become easier to create a cleaner validation flow:

```suggestion
			govy.ForPointer(func(a AnomalyConfigNoData) *string { return a.AlertAfter }).
				WithName("alertAfter").
				Include(govy.New[string](
					govy.Transform(govy.GetSelf[string](), time.ParseDuration).
						Rules(
							rules.DurationPrecision(time.Minute),
							minimalNoDataAlertAfterRule,
							rules.LTE(31*time.Hour*24)),
				)),
```

manifest/v1alpha/slo/validation_test.go
```diff
@@ -443,6 +446,41 @@ func TestValidate_Spec_AnomalyConfig(t *testing.T) {
...
+				ExpectedErrors: []testutils.ExpectedError{
+					{
+						Prop:    "spec.anomalyConfig.noData.alertAfter",
+						Code:    govy.ErrorCodeTransform,
+						Message: `should be greater than or equal to '5m0s'`,
+					},
+				},
+				ExpectedErrorsCount: 1,
+			},
+			"alert after is too long": {
```
nieomylnieja, line 467: you're missing test case for duration precision error

nieomylnieja (changes requested):

manifest/v1alpha/slo/validation.go
nobl9-adam-szymanski: Oh nice, and now validation error is more meaningful:
```
ENCOUNTERED:
        	            	[
        	            	 {
        	            	  "propertyName": "spec.anomalyConfig.noData.alertAfter",
        	            	  "propertyValue": "0s",
        	            	  "errors": [
        	            	   {
        	            	    "error": "should be greater than or equal to '5m0s'",
        	            	    "code": "greater_than_or_equal_to",
        	            	    "description": "should be greater than or equal to '5m0s'"
        	            	   }
        	            	  ]
        	            	 }
        	            	]
```
compared to what it was before:
```
{
        	            	 "property": "spec.anomalyConfig.noData.alertAfter",
        	            	 "code": "transform",
        	            	 "message": "should be greater than or equal to '5m0s'"
        	            	}
```

nieomylnieja (approved):

manifest/v1alpha/slo/validation.go
```diff
@@ -282,8 +282,16 @@ var anomalyConfigValidation = govy.New[AnomalyConfig](
...
-		)),
-)
+			govy.ForPointer(func(a AnomalyConfigNoData) *string { return a.AlertAfter }).
+				WithName("alertAfter").
+				Include(govy.New[string](
+					govy.Transform(govy.GetSelf[string](), time.ParseDuration).
+						Rules(
+							rules.DurationPrecision(time.Minute),
+							minimalNoDataAlertAfterRule,
+							rules.LTE(31*time.Hour*24)),
```
nuusk, line 292: why `minimalNoDataAlertAfterRule` is a variable, and maximum is written inline? 🤔

nuusk (approved):
code looks good. I left a nit pick regarding min-max variable rules

manifest/v1alpha/slo/validation.go
nobl9-adam-szymanski: Thanks, I missed that. Previously (before Mateusz CR), variable `minimalNoDataAlertAfterRule` was used twice. Now it can be removed.

PR #703, opened 2025-05-06
feat: Add send resolution to Service Now

nuusk:
## Motivation

All-clear is not only supported for Pager Duty. [[link]](https://docs.nobl9.com/alerting/alert-methods/pagerduty?_highlight=send#sending-notification-for-a-resolved-alert-to-pagerduty)
We want to enable it for ServiceNow as well. 

## Summary

- Add `SendResolution` field to `ServiceNowAlertMethod`, similarly to `PagerDutyAlertMethod`
- Add validation max 128 characters to `sendResolution.message` (both pagerduty and servicenow) to prevent from any `413` errors from alert methods' APIs. There are no strict limits on the number of characters but this is just a soft limit on our side.

## Release Notes

- `sendResolution` is now supported in Service Now alert method configuration, which enables all-clear mechanism (closing alerts in Service Now after notification cooldown period, with specified `message`)


manifest/v1alpha/alertmethod/validation.go
```diff
@@ -158,6 +158,16 @@ var pagerDutyValidation = govy.New[PagerDutyAlertMethod](
...
+	govy.ForPointer(func(s PagerDutyAlertMethod) *SendResolution { return s.SendResolution }).
+		WithName("sendResolution").
+		Include(sendResolutionValidation),
+)
+
+var sendResolutionValidation = govy.New[SendResolution](
+	govy.ForPointer(func(s SendResolution) *string { return s.Message }).
+		WithName("message").
+		OmitEmpty().
+		Rules(rules.StringMaxLength(128)),
```
nobl9-adam-szymanski, line 170: Do we know that there is AlertMethod for PagerDuty that exceeded that new limit? I mean, that would break applying SLO as a code.

PR #708, opened 2025-05-12
fix: Solve segmentation violation errors for empty APIVersion [PC-16865]

nieomylnieja:
## Summary

Pinned the `go-yaml` version until https://github.com/goccy/go-yaml/pull/698 is released.

## Release Notes

Fixes segmentation violation errors when `apiVersion` was set to an empty string in any of the v1alpha objects.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,9 @@ require (
 	golang.org/x/time v0.11.0
 )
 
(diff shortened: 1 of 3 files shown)


daniel-zelazny (approved):
:rocket:

PR #709, opened 2025-05-12
feat: Allow to configure Email alert method as plain text

nobl9-adam-szymanski:
## Motivation

Users wants to be able to configure Email alert method as plain text (force sending as `content-type: text/plain`).

## Summary

Allow to configure Email alert method to force sending as plain text.

- Sending alert emails as plain text

```yaml
- apiVersion: n9/v1alpha
  kind: AlertMethod
  metadata:
    displayName: Email as plain text
    name: email-as-plain-text
    project: default
  spec:
    description: ""
    email:
      sendAsPlainText: true
      to:
      - test@example.com
```

- Sending alert emails as HTML (default)

```yaml
- apiVersion: n9/v1alpha
  kind: AlertMethod
  metadata:
    displayName: Email as HTML
    name: email-as-plain-text
    project: default
  spec:
    description: ""
    email:
      sendAsPlainText: false
      to:
      - test@example.com
```
or
```yaml
- apiVersion: n9/v1alpha
  kind: AlertMethod
  metadata:
    displayName: Email as HTML
    name: email-as-plain-text
    project: default
  spec:
    description: ""
    email:
      to:
      - test@example.com
```

## Testing

Apply alert method via sloctl.

## Release Notes

Allow to configure Email alert method to force sending alert and anomaly notification as plain text. This change is backward compatible, where existing email alert methods are set as HTML type (default).


nieomylnieja:
Maybe it would be more versatile to allow users to set a custom content type?

PR #748, opened 2025-06-24
feat: Expose v1alpha examples as a package

nieomylnieja:
## Motivation

We have an exhaustive collection of object examples already defined, it would be of great benefit both for our users and us to be able to use access these examples from other projects.
One example of such usage is Terraform Provider, where we could use these examples to write acceptance tests quicker.

## Release Notes

Package `manifest/v1alpha/examples` is now available publicly, it contains en exhaustive list of examples for each manifest Object which can be accessed and freely used.


internal/cmd/examplegen/main.go
```diff
@@ -9,14 +9,14 @@ import (
 
 	"github.com/goccy/go-yaml"
 
-	v1alphaExamples "github.com/nobl9/nobl9-go/internal/manifest/v1alpha/examples"
 	"github.com/nobl9/nobl9-go/internal/pathutils"
 	"github.com/nobl9/nobl9-go/manifest"
+	v1alphaExamples2 "github.com/nobl9/nobl9-go/manifest/v1alpha/examples"
```
skrolikiewicz, line 14: Is `..2` intended to be here? It looks like it might be a placeholder or a leftover.

PR #756, opened 2025-07-04
feat: Add Replay API support

nieomylnieja:
## Motivation

Replay clients currently maintain separate HTTP requests and models. A versioned SDK interface provides one public contract for managing Replays and reduces drift between consumers.

## Summary

- Added Replay V1 methods to run Replays, list active requests, inspect status, request cancellation, delete queued requests, and check availability.
- Added validated requests and typed response values for Replay types, statuses, cancellation states, sources, durations, and availability reasons.
- Preserved RFC3339 `createdAt` values and unknown availability reasons returned by the Nobl9 platform.
- Required availability requests to select either an SLO or a data source and rejected non-positive or overflowing durations without reintroducing the former 30-day SDK limit.
- Replaced the legacy `sdk/models` Replay API with `sdk/endpoints/replay/v1`.

## Related changes

- [sloctl #482](https://github.com/nobl9/sloctl/pull/482)

## Testing

- Added end-to-end coverage for availability by SLO and data source, `timeRange.startDate`, source-SLO objective mapping, and `Run`, `List`, `GetStatus`, `Cancel`, and `Delete` flows, including queued, in-progress, canceled, and completed states plus RFC3339 `createdAt`.
- Added unit coverage for request validation and serialization, selector modes, duration boundaries and overflow, status conversion, and unknown availability reasons.

## Release Notes

Added a versioned Replay API to start Replays, inspect status and availability, list active requests, request cancellation, and delete queued requests. Replay types now live under `sdk/endpoints/replay/v1`.

## Breaking Changes

Removed the `sdk/models` package, including its legacy Replay types and helpers. Consumers must migrate to `sdk/endpoints/replay/v1`. `models.Replay` is replaced by `v1.RunRequest`, acronym fields use `SLO`, and the unsupported `EndDate` field has no replacement.


sdk/endpoints/replay/v1/response.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type ReplayListItem struct {
+	SLO            string `json:"slo,omitempty"`
+	Project        string `json:"project"`
+	ElapsedTime    string `json:"elapsedTime,omitempty"`
+	RetrievedScope string `json:"retrievedScope,omitempty"`
+	RetrievedFrom  string `json:"retrievedFrom,omitempty"`
+	Status         string `json:"status"`
+	Cancellation   string `json:"cancellation,omitempty"`
```
ditrytus: This field also has a predefined list of viariants.

sdk/endpoints/replay/v1/endpoints.go
```diff
@@ -0,0 +1,186 @@
...
+	resp, err := e.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = resp.Body.Close() }()
+	return nil
+}
+
+func (e endpoints) Delete(ctx context.Context, params DeleteRequest) (err error) {
+	return e.deleteReplay(ctx, params)
```
ditrytus: `deleteReplay` can be inlined here. It is not used anywhere else and `Delete` is now just a passthrough function.

sdk/endpoints/replay/v1/request.go
```diff
@@ -0,0 +1,244 @@
...
+type GetAvailabilityRequest struct {
+	Project           string
+	DataSourceProject string
+	DataSource        string
+	DataSourceKind    string
+	SLOName           string
+	Type              string
+	DurationUnit      string
+	DurationValue     int
+}
```
ditrytus, line 169: Doesn't this struct don't require struct tags? Maybe if this one doesn't the others don't as well?

ditrytus (changes requested):
There is a regression in relation to this change: https://github.com/nobl9/nobl9-go/pull/845 (30 day limit enforced in SDK).

sdk/endpoints/replay/v1/response.go
nieomylnieja: @ditrytus Added a `ReplayListStatus...` constant group for `ReplayListItem.Status`, mirroring the known `failed` and `completed` status variants without changing the response field type.

sdk/endpoints/replay/v1/endpoints.go
nieomylnieja: @ditrytus Inlined the delete request implementation into `Delete` and removed the unused `deleteReplay` helper.

sdk/endpoints/replay/v1/response.go
nieomylnieja: @ditrytus Follow-up: changed `ReplayStatus.Status`, `ReplayStatus.Cancellation`, `ReplayListItem.Status`, and `ReplayListItem.Cancellation` to named string enum types. Added typed constants for the known status variants and the known cancellation variant (`blocked`).

nieomylnieja: @ditrytus Follow-up after checking the n9 implementation: `source` now uses `ReplaySource` (`user`, `error_budget_adjustment`), `replayType`/availability `type` use `ReplayType` (`reimport_and_recalculation`, `recalculation`), `ReplayListItem.Status` uses the list status set (`unknown`, `queued`, `in progress`, `completed`, `failed`, `canceled`), and `ReplayStatus.Cancellation`/list cancellation use the full cancellation set (`possible`, `blocked`, `requested`, `denied`, `done`). I also added the fixed availability reasons currently defined in n9.

ditrytus (changes requested):

sdk/endpoints/replay/v1/response.go
ditrytus: Cancellation is fine, but status needs to be changed. I'll describe it in a separate comment: https://github.com/nobl9/nobl9-go/pull/756#discussion_r3497820299

sdk/endpoints/replay/v1/request.go
ditrytus: What about this question?

sdk/endpoints/replay/v1/response.go
nieomylnieja: @ditrytus Follow-up handled in the status thread: `ReplayProcessStatus` is now trimmed to the coarse n9 public set and `ReplayListStatus` aliases it. Cancellation remains aligned with n9: `possible`, `blocked`, `requested`, `denied`, and `done`.

sdk/endpoints/replay/v1/request.go
nieomylnieja: this one is only used as a intermediate input object for the users, it is not used as the actual payload, that's why it doesn't have the tags

sdk/endpoints/replay/v1/response.go
```diff
@@ -9,21 +9,21 @@ type ReplayWithStatus struct {
...
-	Unit         string `json:"unit"`
-	StartTime    string `json:"startTime"`
-	EndTime      string `json:"endTime,omitempty"`
-	Value        int    `json:"value"`
+	Source       ReplaySource             `json:"source"`
+	Status       ReplayProcessStatus      `json:"status"`
+	Cancellation ReplayCancellationStatus `json:"cancellation"`
+	CanceledBy   string                   `json:"canceledBy,omitempty"`
+	TriggeredBy  string                   `json:"triggeredBy"`
+	Unit         string                   `json:"unit"`
```
ditrytus: This type can be changed from `string` to `DurationUnit` you just created.

sdk/endpoints/replay/v1/response.go
```diff
@@ -9,21 +9,21 @@ type ReplayWithStatus struct {
...
-	Status       string `json:"status"`
-	Cancellation string `json:"cancellation"`
-	CanceledBy   string `json:"canceledBy,omitempty"`
-	TriggeredBy  string `json:"triggeredBy"`
-	Unit         string `json:"unit"`
-	StartTime    string `json:"startTime"`
-	EndTime      string `json:"endTime,omitempty"`
-	Value        int    `json:"value"`
+	Source       ReplaySource             `json:"source"`
+	Status       ReplayProcessStatus      `json:"status"`
```
ditrytus: This is the response where ingest will not send the values that are in `ReplayProcessStatus` enum, but the fine grained statuses that should be private. I would either change it back to `string` or update ingest.

ditrytus (changes requested):

sdk/endpoints/replay/v1/response.go
nieomylnieja: I'll update the internal code path then

sdk/endpoints/replay/v1/response.go
nieomylnieja: Changed `ReplayStatus.Unit` to `DurationUnit` and kept `ToProcessStatus` converting it back to `string` for manifest process status. Added response unmarshal coverage for `Hour`.

sdk/endpoints/replay/v1/response.go
nieomylnieja: Kept the SDK enum on the coarse public status set (unknown, queued, in progress, completed, failed, canceled) and handled the producer-side mapping separately. I’m intentionally not linking that work from this public SDK thread.

sdk/endpoints/replay/v1/response.go
nieomylnieja: Addressed in the current branch: ReplayStatus.Unit now uses DurationUnit.

sdk/endpoints/replay/v1/response.go
nieomylnieja: Changed ReplayStatus.Status back to string so the single-status endpoint preserves its fine-grained, open-ended values. ReplayListItem.Status remains the typed ReplayListStatus because the list endpoint returns the separate coarse status contract; the response test now covers fetching_historical_data explicitly.

sdk/endpoints/replay/v1/request.go
```diff
@@ -0,0 +1,427 @@
...
+			_, err := manifest.ParseKind(kind)
+			return err
+		}).WithErrorCode(rules.ErrorCodeOneOf)),
+	govy.For(func(r GetAvailabilityRequest) GetAvailabilityRequest { return r }).
+		Rules(rules.MutuallyExclusive(false, map[string]func(GetAvailabilityRequest) any{
+			"sloName": func(r GetAvailabilityRequest) any { return r.SLOName },
+			"dataSource": func(r GetAvailabilityRequest) any {
+				return r.DataSourceProject != "" || r.DataSource != "" || r.DataSourceKind != ""
+			},
+		})),
```
matkaras, line 330: issue: The SLO details UI still calls this endpoint with both sloName and the data source query parameters. GetAvailabilityRequest.Validate() treats these selector groups as mutually exclusive, and the new test explicitly expects HTTP 400 for that request. As a result, the existing availability check for a non-composite SLO will fail. We should either update the frontend caller in this PR to send only sloName, or preserve the previous server behavior where sloName takes precedence.


example:  `https://dev-purple.nobl9.dev/api/internal/timemachine/availability?dataSourceProject=splunk-direct&dataSource=splunk-direct&dataSourceKind=Direct&sloName=aaaaasasasasdadd1&`

tests/replay_v1_test.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/nobl9/nobl9-go/manifest/v1alpha"
+	v1alphaDirect "github.com/nobl9/nobl9-go/manifest/v1alpha/direct"
+	v1alphaService "github.com/nobl9/nobl9-go/manifest/v1alpha/service"
+	v1alphaSLO "github.com/nobl9/nobl9-go/manifest/v1alpha/slo"
+	"github.com/nobl9/nobl9-go/sdk"
+	replayV1 "github.com/nobl9/nobl9-go/sdk/endpoints/replay/v1"
+	"github.com/nobl9/nobl9-go/tests/e2etestutils"
+)
+
+func Test_Replay_V1(t *testing.T) {
```
matkaras, line 25: nitpick: it would be nice to add some comments or helper methods to make it easier to follow

sdk/endpoints/replay/v1/request.go
```diff
@@ -0,0 +1,427 @@
...
+// Duration defines how far back a replay should retrieve data.
+type Duration struct {
+	Unit  DurationUnit `json:"unit"`
+	Value int          `json:"value"`
+}
+
+// TimeRange defines the earliest point from which a replay should retrieve data.
+type TimeRange struct {
+	StartDate time.Time `json:"startDate,omitzero"`
+}
```
matkaras: issue: this is not time range

sdk/endpoints/replay/v1/request.go
nieomylnieja: Kept the SDK validation because SLO and data-source selectors are separate request modes.
  The companion platform change now sends only `project` and `sloName` for the SLO flow.
  The API also keeps `sloName` precedence for legacy mixed queries.
  This preserves existing callers while preventing new typed callers from sending ambiguous
  requests.
  Coverage verifies the SLO-only request, legacy precedence, and SDK rejection of mixed selectors.

tests/replay_v1_test.go
nieomylnieja: Split the e2e test into focused helpers for the no-existing-Replay,
  availability, cancellation, and queue-lifecycle scenarios.
  I also extracted shared status polling and cleanup.
  The top-level test now only creates fixtures and sequences those scenarios.

sdk/endpoints/replay/v1/request.go
nieomylnieja: The wire contract is `timeRange.startDate`; it has no end-date field.
Request processing time is the effective end,
so `TimeRange` still matches the JSON object rather than representing a closed interval.
I documented this semantic,
added exact `timeRange.startDate` serialization coverage,
and exercised it in the cancellation e2e scenario.

matkaras (approved):

ditrytus (approved):

PR #774, opened 2025-08-04
fix: correct validation for composite SLOs

nieomylnieja:
## Release Notes

Hardened validation for instances when `rawMetric` or `countMetrics` was provided with `composite` section in the `spec.objective` definition.

### DIFF
--- a/manifest/v1alpha/slo/validation.go
+++ b/manifest/v1alpha/slo/validation.go
@@ -415,8 +415,21 @@ var specValidationComposite = govy.New[Spec](
 		IncludeForEach(govy.New[Objective](
 			govy.For(func(o Objective) ObjectiveBase { return o.ObjectiveBase }).
 				Include(objectiveBaseValidation),
+			govy.For(func(o Objective) *RawMetricSpec { return o.RawMetric }).
+				WithName("rawMetric").
+				Rules(rules.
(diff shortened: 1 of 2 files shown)


manifest/v1alpha/slo/validation.go
```diff
@@ -436,7 +449,9 @@ var specValidationComposite = govy.New[Spec](
...
 				)),
 		)),
-).When(
-	func(s Spec) bool { return s.HasCompositeObjectives() },
-	govy.WhenDescription("at least one composite objective is defined"),
-)
+).
+	When(
+		func(s Spec) bool { return s.HasCompositeObjectives() },
+		govy.WhenDescription("at least one composite objective is defined"),
```
nikodemrafalski, line 455: I see it's older issue, but a bit misleading since composite SLO can have exactly ONE composite objective

nikodemrafalski (approved):

manifest/v1alpha/slo/validation.go
nieomylnieja: that's a "when" description, it informs you (in the validation plan) when the given rules are taking effect, so it's perfectly fine ;) either way, we're not using it YET anywhere other than nobl9 language server

PR #776, opened 2025-08-08
feat: add SumoLogic to data retrieval max duration mappings

nieomylnieja:
## Release Notes

SumoLogic now supports Replay and SLI Analysis features.


ditrytus (approved):
I am approving because the changes look sound, but I am not knowledgeable enough when it comes to the nobl9-go project to be sure that these are all the changes that need to be done.

PR #806, opened 2025-10-02
feat: add support for removing computed fields from manifest objects [PC-17898]

nieomylnieja:
## Motivation

Currently when fetching Nobl9 Objects from the API users have no way of removing computed fields from the response.
These fields are useful in many scenarios, but when it comes to managing Nobl9 configuration as a code, they are problematic, since control version systems will detect changes on them each time the object is fetched.

## Summary

Introduce a new utility function `RemoveComputedFieldsFromObjects` in the SDK to remove fields tagged with `nobl9:"computed"` from manifest objects. This function recursively traverses the object structure and clears computed fields, ensuring that only user-defined fields remain.

Additionally, a comprehensive test suite has been added to validate the functionality across various scenarios, including objects with nested structures, multiple objects, and objects without computed fields.

## Testing

Added unit tests for the method.

## Release Notes

Added `sdk.RemoveComputedFieldsFromObjects` function allows practitioners to remove computed fields from `manifest.Object`.


manifest/v1alpha/slo/slo.go
```diff
@@ -52,10 +52,10 @@ type Spec struct {
...
 	// Read-only field.
-	CreatedAt string `json:"createdAt,omitempty"`
+	CreatedAt string `json:"createdAt,omitempty" nobl9:"computed"`
 	// CreatedBy is the id of the user who first created the SLO.
 	// Read-only field.
-	CreatedBy string `json:"createdBy,omitempty"`
+	CreatedBy string `json:"createdBy,omitempty" nobl9:"computed"`
 	// Deprecated: this implementation of Composite will be removed and replaced with SLO.Spec.Objectives.Composite.
 	Composite     *Composite     `json:"composite,omitempty"`
 	AnomalyConfig *AnomalyConfig `json:"anomalyConfig,omitempty"`
```
nobl9-adam-szymanski, line 61:
```
type ObjectiveBase struct {
    ...
	NameChanged bool     `json:"-"`
}
```
Is NameChanged treated as computed?

nobl9-adam-szymanski (approved):
Solid work. I checked v1alpha objects and I didn't find any other fields that we should mark as computed. I tested it and it works as expected for kind SLO.

PR #893, opened 2026-03-09
feat: PI-1274 Add SendResolution support to TeamsAlertMethod

bsingern9:
## Summary
- Adds `SendResolution` field to `TeamsAlertMethod` struct, enabling all-clear (resolution) notifications for MS Teams alert methods
- Adds validation for the new field, reusing the existing `sendResolutionValidation` (same pattern as PagerDuty and ServiceNow)

## Context
JIRA: PI-1274

Customers want to receive "alert resolved" notifications via MS Teams when alerts are resolved or canceled. This change adds the manifest-level support so that the `n9` backend can wire `SendResolution` for Teams alert methods.

## Test plan
- [x] Existing validation tests continue to pass
- [x] New field follows the same pattern as PagerDuty/ServiceNow SendResolution
- [ ] Companion PR in n9 repo wires the backend notification logic

### DIFF
--- a/manifest/v1alpha/alertmethod/alert_method.go
+++ b/manifest/v1alpha/alertmethod/alert_method.go
@@ -135,7 +135,8 @@ type JiraAlertMethod struct {
 
 // TeamsAlertMethod represents a set of properties required create Microsoft Teams notifications.
 type TeamsAlertMethod struct {
-	URL string `json:"url"`
+	URL            string          `json:"url"`
+	SendResolution *SendResolution `json:"sendResolution,omitempty"`
 }
 
 // EmailAlertMethod represents a set of properties required to send an email.
--- a/manifest/v1alpha/alertmethod/validation.go
+++ b/manifest/v1alpha/alertmethod/validation.go
@@ -270,6 +270,9 @@ var teamsValidation = govy.New[TeamsAlertMethod](
 				return nil
 			}),
 		),
+	govy.ForPointer(func(s TeamsAlertMethod) *SendResolution { return s.SendResolution }).
+		WithName("sendResolution").
+		Include(sendResolutionValidation),
 ).When(
 	func(v TeamsAlertMethod) bool { return !isHiddenValue(v.URL) },
 	govy.WhenDescriptionf("is empty or equal to '%s'", v1alpha.HiddenValue),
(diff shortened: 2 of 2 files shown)


n9-machine-user (changes requested):
**Scope: XS — appropriate.** Two issues need fixing before this can go forward.

---

**1. Validation scope bug (inline comment on validation.go:273-275)**

The `sendResolution` validation ends up inside the `.When(!isHiddenValue(v.URL))` block because it's added to the same `govy.New[TeamsAlertMethod](...)` that the `When()` is chained on. When URL is empty or hidden, the entire validator is skipped — including `sendResolution`. PagerDuty and ServiceNow don't have this problem because their `sendResolution` validation is at the top-level validator without any conditional. This needs to be restructured.

---

**2. Example not updated (`manifest/v1alpha/examples/alert_method.go`)**

Both PagerDuty and ServiceNow set `SendResolution` in their example objects. The Teams case was not updated and still only sets `URL`. For consistency and discoverability, the Teams example should include `SendResolution` too.

---

**3. No new tests (minor)**

`validation_teams_test.go` was not modified. Existing PagerDuty/ServiceNow tests also lack `sendResolution` coverage, so this is consistent with the codebase norm — but since this PR is explicitly adding the field, at minimum a "passes with sendResolution set" case would be appropriate.

n9-machine-user (approved):
Validation bug is fixed correctly. `When()` is now field-level on `url` only — `sendResolution` validation runs unconditionally regardless of URL state, matching the `serviceNowValidation` structure exactly.

**Remaining gaps for human reviewers to consider:**

1. `manifest/v1alpha/examples/alert_method.go` — Teams example still only sets `URL`, no `SendResolution`. PagerDuty and ServiceNow both include it. This means the e2e test fixture for Teams won't exercise the new field. Low risk, but worth a note — may want to add it once the companion n9 backend PR lands (so e2e tests don't hit an unsupported field).

2. `validation_teams_test.go` — no new test cases for `sendResolution`. Consistent with the existing PagerDuty/ServiceNow tests which also lack sendResolution coverage, so this is more of a codebase-wide gap than a blocker here.

**Scope: XS.** No migrations, clean rollback, correct implementation.

PR #914, opened 2026-05-06
feat: support custom CA bundle via Config.CACertFile

diliuskh:
## Motivation

  On a corporate macOS laptop with an MDM-installed trust profile, every Go-built `sloctl` invocation (and any direct `nobl9-go` SDK consumer) fails at the IDP token step:

  ```
  Post "https://accounts.nobl9.com/oauth2/.../v1/token":
    tls: failed to verify certificate: x509: OSStatus -26276
  ```

  The chain Nobl9 presents (`*.accounts.nobl9.com` → `Amazon RSA 2048 M01` → `Amazon Root CA 1`) is genuine and is accepted by every other tool on the same machine:

  - `openssl s_client -connect accounts.nobl9.com:443`
  - `security verify-cert` against `/Library/Keychains/System.keychain`
  - `curl https://accounts.nobl9.com/`

  It is rejected only by Go's `crypto/x509` on Darwin, which delegates chain validation to `Security.SecTrustEvaluateWithError`. The MDM trust profile (or one of the overlapping policy hooks it installs) causes
  the platform call to return `-26276` even though the system's default policy accepts the same chain.

  The standard escape hatches do not help, because Go on Darwin always routes through Security.framework when `tls.Config.RootCAs == nil`:

  - `SSL_CERT_FILE` / `SSL_CERT_DIR` — ignored on Darwin in this code path
  - `GODEBUG=x509usefallbackroots=1` — only fires after a *load* failure, not an `Evaluate` rejection
  - `CGO_ENABLED=0` — irrelevant; `crypto/x509/root_darwin.go` always uses Security.framework

  The supported, minimally invasive workaround is to give the SDK a way to declare its own trust store. When `tls.Config.RootCAs` is set explicitly, Go uses its pure-Go `x509.Verify` and ignores the platform
  verifier entirely. That is the entire fix.

  ## Summary

  Adds an opt-in `Config.CACertFile` field that loads a PEM-encoded CA bundle and injects it as `tls.Config.RootCAs` into both authentication paths used by `Client`:

  - the Okta IDP token request transport (`okta.go`)
  - the inner `http.Client` used by `credentialsStore.RoundTrip` for every authenticated Nobl9 API call

  Three equivalent ways to set it (matches the existing `ClientID` / `OktaOrgURL` plumbing):

  - TOML: `caCertFile = "/path/to/ca-bundle.pem"` under `[contexts.<name>]`
  - Env: `NOBL9_SDK_CA_CERT_FILE=/path/to/ca-bundle.pem` (or whatever prefix the consumer set via `ConfigOptionEnvPrefix`, e.g. `SLOCTL_CA_CERT_FILE` in sloctl)
  - Programmatic: `cfg.CACertFile = "/path/to/ca-bundle.pem"` on the returned `*sdk.Config` before calling `NewClient`

  When unset, behavior is unchanged: clients fall back to `http.DefaultTransport` with the platform default trust store.

  Implementation details:

  - New `sdk/tls.go` with `newCustomCATransport(path) (http.RoundTripper, error)`. Returns nil when path is empty (caller falls back to default), otherwise reads the PEM, builds an `x509.CertPool`, clones
  `http.DefaultTransport`, and sets `TLSClientConfig.RootCAs` (with `MinVersion: TLS 1.2`).
  - `sdk/config.go`: `Config.CACertFile string` and `ContextConfig.CACertFile` with `toml`/`json`/`env` tags. Wired through `resolveContextConfig`.
  - `sdk/credentials.go`: `newCredentials` now returns `(*credentialsStore, error)` so it can surface PEM parse failures up front. The package-level `cleanCredentialsHTTPClient` global is replaced by a per-store
  `apiHTTPClient *http.Client` field carrying the configured transport.
  - `sdk/okta.go`: `newOktaClient(getEndpoint, customTransport)` accepts the transport and threads it into `newRetryableHTTPClient`.
  - `sdk/client.go`: handles the new error from `newCredentials`.

  ## Related changes

  None. This is a self-contained SDK change. Once it lands and is tagged in a release, downstream consumers (sloctl, terraform-provider-nobl9, custom integrations) get the fix via a normal `go get` bump with no
  client-side code change required.

  ## Testing

  ### Unit tests

  - `sdk/tls_test.go::TestNewCustomCATransport` covers the helper directly: empty path returns nil, missing file returns wrapped error, non-PEM input returns parse error, and a TLS handshake against an
  `httptest.NewTLSServer` whose leaf cert is the only entry in the bundle succeeds.
  - `sdk/tls_test.go::TestNewClient_CACertFile` covers the construction path:
    - empty `CACertFile` leaves `apiHTTPClient.Transport` nil (matches pre-change `cleanCredentialsHTTPClient` behavior)
    - valid bundle wires `RootCAs` into `credentialsStore.apiHTTPClient.Transport`, which is what guarantees the pure-Go x509 verifier path is taken
  - `sdk/config_test.go`: extended `TestReadConfig_FromFullConfigFile` and `TestReadConfig_EnvVariablesFull` to cover both TOML and env-variable loading of the new field. Fixtures in `sdk/test_data/config/`
  updated.

  Full unit suite passes with `-race -cover` (matches the `tests.yml` CI invocation).

  ### End-to-end

  Validated on a corporate macOS laptop with this branch wired into `sloctl` via a local `replace` directive, then setting `caCertFile` in `~/.config/nobl9/config.toml` to a bundle composed of `/etc/ssl/cert.pem`
   plus the company-issued CAs exported from the System keychain:

  ```
  $ sloctl get projects
  - apiVersion: n9/v1alpha
    kind: Project
    metadata:
      name: <real-project>
      ...
  ```

  The same binary without `caCertFile` set still reproduces the original `OSStatus -26276` failure on the same machine, confirming the change is opt-in and additive.

  ### CI parity

  All targets from `.github/workflows/checks.yml` and `.github/workflows/vulns.yml` pass locally:

  - `make check/format`
  - `make check/vet`
  - `make check/lint` — `0 issues`
  - `make check/spell` — `0 issues` (`wincrypt` added to the dictionary for the platform-verifier explanation)
  - `make check/trailing`
  - `make check/markdown`
  - `make check/generate`
  - `make check/vulns` — `0 vulnerabilities`

  ## Out of scope (follow-up candidates)

  - `sdk/reader.go::httpClientFactory` (used by `ReadObjectsFromSources` for fetching YAML definitions over HTTP) is a package-global with no transport, separate from the auth path. In environments where users
  fetch their own definitions over internal-CA HTTPS, that path would still hit the same root cause. A follow-up PR plumbing `Config.CACertFile` (or a request-scoped option) into `httpClientFactory` would close
  the gap. Left out of this PR to keep the diff focused on the auth path that blocks every command, not just `apply -f https://...`.
  - No `InsecureSkipVerify` toggle. The escape hatch is "trust this bundle", not "trust nobody". If a use case for skipping verification entirely materializes, that should be a separate PR with its own
  justification.

  ## Release Notes

  Add an opt-in `Config.CACertFile` (also bound to `NOBL9_SDK_CA_CERT_FILE` and the per-context `caCertFile` TOML key) for loading a custom CA bundle. When set, the SDK uses Go's pure-Go x509 verifier with the
  supplied roots instead of the platform verifier, which fixes `tls: failed to verify certificate: x509: OSStatus -26276` on macOS hosts whose MDM trust profile interferes with `SecTrustEvaluateWithError`.


sdk/tls.go
```diff
@@ -0,0 +1,42 @@
...
+// laptops with MDM-installed trust profiles that interfere with SecTrustEvaluateWithError.
+func newCustomCATransport(caCertFile string) (http.RoundTripper, error) {
+	if caCertFile == "" {
+		return nil, nil
+	}
+	pemBytes, err := os.ReadFile(caCertFile) // nolint: gosec
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA bundle %q: %w", caCertFile, err)
+	}
+	pool := x509.NewCertPool()
```
nieomylnieja: If we're extending the certificates, as you documented in `ContextConfig.CACertFile`, we should use `SystemCertPool`:

```suggestion
	pool, err := x509.SystemCertPool()
	if err != nil {
		return nil, fmt.Errorf("failed to load system cert pool: %w", err)
	}
```

nieomylnieja (changes requested):
hey @diliuskh, thanks for the contribution!
I've one thing to verify, see the comment for review. In addition I've made some adjustments to the submitted code (see the patch below), please apply them and verify the change still works for you!
Once that's done I don't see any issues around merging it.
Once it's merged, I will release a new SDK version and use it in sloctl, and finally, release new sloctl version.

```patch
diff --git a/sdk/config.go b/sdk/config.go
index 7fb70b3..b008501 100644
--- a/sdk/config.go
+++ b/sdk/config.go
@@ -93,8 +93,12 @@ type Config struct {
 	Organization string
 	// Timeout is the timeout duration of each HTTP request run against the API.
 	Timeout time.Duration
-	// CACertFile is an optional path to a PEM-encoded bundle of CA certificates.
-	// See [ContextConfig.CACertFile] for full semantics.
+	// CACertFile is an optional path to a PEM-encoded bundle of CA certificates that will be
+	// trusted in addition to the system cert pool when establishing TLS connections to
+	// the Nobl9 API and IDP. When set it forces the use of Go's native x509 verifier
+	// instead of the platform verifier (Security.framework on macOS, wincrypt on Windows),
+	// which is the supported escape hatch for environments where the platform verifier
+	// rejects a chain that the Go verifier accepts.
 	CACertFile string
 	// FilesPromptEnabled is a flag that enables a prompt for applying/deleting large numbers of files.
 	// It is sloctl exclusive.
@@ -145,13 +149,7 @@ type ContextConfig struct {
 	DisableOkta    *bool          `toml:"disableOkta,omitempty" json:"disableOkta,omitempty" env:"DISABLE_OKTA"`
 	Organization   string         `toml:"organization,omitempty" json:"organization,omitempty" env:"ORGANIZATION"`
 	Timeout        *time.Duration `toml:"timeout,omitempty" json:"timeout,omitempty" env:"TIMEOUT"`
-	// CACertFile is an optional path to a PEM-encoded bundle of CA certificates that will be
-	// trusted in addition to the default Mozilla roots when establishing TLS connections to
-	// the Nobl9 API and IDP. When set it forces the use of Go's pure-Go x509 verifier
-	// instead of the platform verifier (Security.framework on macOS, wincrypt on Windows),
-	// which is the supported escape hatch for environments where the platform verifier
-	// rejects a chain that the Go verifier accepts.
-	CACertFile string `toml:"caCertFile,omitempty" json:"caCertFile,omitempty" env:"CA_CERT_FILE"`
+	CACertFile     string         `toml:"caCertFile,omitempty" json:"caCertFile,omitempty" env:"CA_CERT_FILE"`
 }
 
 // ConfigOption conveys extra configuration details for [ReadConfig] function.
diff --git a/sdk/credentials.go b/sdk/credentials.go
index ea2763f..67f895c 100644
--- a/sdk/credentials.go
+++ b/sdk/credentials.go
@@ -25,13 +25,16 @@ type accessTokenProvider interface {
 type accessTokenPostRequestHook = func(token string) error
 
 func newCredentials(config *Config) (*credentialsStore, error) {
-	customTransport, err := newCustomCATransport(config.CACertFile)
-	if err != nil {
-		return nil, err
+	transport := http.RoundTripper(http.DefaultTransport)
+	if config.CACertFile != "" {
+		customCATransport, err := newCustomCATransport(config.CACertFile)
+		if err != nil {
+			return nil, err
+		}
+		transport = customCATransport
 	}
-	apiClient := newAPIClient(customTransport)
 	if config.DisableOkta {
-		return &credentialsStore{config: config, apiHTTPClient: apiClient}, nil
+		return &credentialsStore{config: config, apiTransport: transport}, nil
 	}
 	authServerURL := oktaAuthServerURL(config.OktaOrgURL, config.OktaAuthServer)
 	return &credentialsStore{
@@ -39,22 +42,12 @@ func newCredentials(config *Config) (*credentialsStore, error) {
 		tokenParser: newJWTParser(authServerURL.String()),
 		tokenProvider: newOktaClient(func() string {
 			return oktaTokenEndpoint(authServerURL).String()
-		}, customTransport),
+		}, transport),
 		postRequestHook: config.saveAccessToken,
-		apiHTTPClient:   apiClient,
+		apiTransport:    transport,
 	}, nil
 }
 
-// newAPIClient returns the inner [http.Client] used by [credentialsStore.RoundTrip]
-// to issue authenticated requests to the Nobl9 API.
-// It applies a custom transport (with extended trust roots) when one is provided.
-func newAPIClient(rt http.RoundTripper) *http.Client {
-	if rt == nil {
-		return &http.Client{}
-	}
-	return &http.Client{Transport: rt}
-}
-
 // credentialsStore stores and manages IDP app credentials and claims.
 // It governs access token life cycle, providing means of refreshing it
 // and exposing claims delivered with the token.
@@ -72,11 +65,11 @@ type credentialsStore struct {
 	tokenType tokenType
 	claims    *jwtClaims
 
-	HTTP *http.Client
-	// apiHTTPClient is the inner client used by [credentialsStore.RoundTrip].
-	// It carries the optional custom TLS transport so that authenticated requests
+	// apiTransport is the final transport used by [credentialsStore.RoundTrip]
+	// after it refreshes credentials and mutates the request headers.
+	// It carries the optional custom TLS transport so authenticated requests
 	// to the Nobl9 API observe [Config.CACertFile] just like the IDP token request.
-	apiHTTPClient *http.Client
+	apiTransport http.RoundTripper
 	// tokenParser is used to verify the token and its claims.
 	tokenParser accessTokenParser
 	// tokenProvider is used to provide an access token.
@@ -157,10 +150,10 @@ func (c *credentialsStore) GetUserID(ctx context.Context) (string, error) {
 // It will wrap any errors returned from [credentialsStore.refreshAccessToken]
 // in [httpNonRetryableError] to ensure the request is not retried by the wrapping client.
 //
-// The inner [http.Client] is constructed once in [newCredentials] and may carry a
+// The final [http.RoundTripper] is constructed once in [newCredentials] and may carry a
 // custom transport with extended trust roots when [Config.CACertFile] is set.
-// Request middleware in Go is clunky and requires chaining multiple HTTP clients;
-// timeouts and retries are handled by the predecessors of this one.
+// This keeps a single outer [http.Client] while still allowing request middleware
+// to refresh credentials before delegating to the actual transport.
 func (c *credentialsStore) RoundTrip(req *http.Request) (*http.Response, error) {
 	tokenUpdated, err := c.refreshAccessToken(req.Context())
 	if err != nil {
@@ -169,7 +162,11 @@ func (c *credentialsStore) RoundTrip(req *http.Request) (*http.Response, error)
 	if _, authHeaderSet := req.Header[HeaderAuthorization]; tokenUpdated || !authHeaderSet {
 		c.setAuthorizationHeader(req)
 	}
-	return c.apiHTTPClient.Do(req)
+	apiTransport := c.apiTransport
+	if apiTransport == nil {
+		apiTransport = http.DefaultTransport
+	}
+	return apiTransport.RoundTrip(req)
 }
 
 // setAuthorizationHeader sets an authorization header which should be included
diff --git a/sdk/okta.go b/sdk/okta.go
index e9a66da..8d9b1b9 100644
--- a/sdk/okta.go
+++ b/sdk/okta.go
@@ -37,11 +37,9 @@ type oktaClient struct {
 }
 
 // newOktaClient returns a client that talks to the Okta IDP token endpoint.
-// When customTransport is non-nil it is used as the inner [http.RoundTripper] of the
-// retryable client, which is how [Config.CACertFile] reaches the IDP request path.
-func newOktaClient(getTokenEndpoint getTokenEndpointFunc, customTransport http.RoundTripper) *oktaClient {
+func newOktaClient(getTokenEndpoint getTokenEndpointFunc, transport http.RoundTripper) *oktaClient {
 	return &oktaClient{
-		HTTP:             newRetryableHTTPClient(oktaRequestTimeout, customTransport),
+		HTTP:             newRetryableHTTPClient(oktaRequestTimeout, transport),
 		getTokenEndpoint: getTokenEndpoint,
 	}
 }
diff --git a/sdk/tls.go b/sdk/tls.go
index 410c69f..26963ee 100644
--- a/sdk/tls.go
+++ b/sdk/tls.go
@@ -3,16 +3,16 @@ package sdk
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
 )
 
 // newCustomCATransport returns an [http.RoundTripper] that trusts the certificates
-// loaded from the PEM file at caCertFile in addition to clearing any default trust.
-// When caCertFile is empty it returns nil so callers fall back to [http.DefaultTransport].
+// loaded from the PEM file at caCertFile in addition to the system cert pool.
 //
-// Setting [tls.Config.RootCAs] forces Go's pure-Go [crypto/x509] verifier to be used
+// Setting [tls.Config.RootCAs] forces Go's native [crypto/x509] verifier to be used
 // instead of the platform verifier (Security.framework on macOS, wincrypt on Windows).
 // This is the supported escape hatch for environments where the platform verifier
 // rejects a chain that Go's verifier (and the loaded roots) accept - e.g. corporate
@@ -23,15 +23,18 @@ func newCustomCATransport(caCertFile string) (http.RoundTripper, error) {
 	}
 	pemBytes, err := os.ReadFile(caCertFile) // nolint: gosec
 	if err != nil {
-		return nil, fmt.Errorf("failed to read CA bundle %q: %w", caCertFile, err)
+		return nil, fmt.Errorf("failed to read CA bundle from %q: %w", caCertFile, err)
+	}
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load system cert pool: %w", err)
 	}
-	pool := x509.NewCertPool()
 	if !pool.AppendCertsFromPEM(pemBytes) {
 		return nil, fmt.Errorf("failed to parse any PEM certificates from %q", caCertFile)
 	}
 	transport, ok := http.DefaultTransport.(*http.Transport)
 	if !ok {
-		return nil, fmt.Errorf("http.DefaultTransport is not *http.Transport")
+		return nil, errors.New("http.DefaultTransport is not *http.Transport")
 	}
 	cloned := transport.Clone()
 	cloned.TLSClientConfig = &tls.Config{
diff --git a/sdk/tls_test.go b/sdk/tls_test.go
index 08a84ad..cfaad04 100644
--- a/sdk/tls_test.go
+++ b/sdk/tls_test.go
@@ -2,6 +2,7 @@ package sdk
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/pem"
 	"net/http"
 	"net/http/httptest"
@@ -60,12 +61,35 @@ func TestNewCustomCATransport(t *testing.T) {
 		t.Cleanup(func() { _ = resp.Body.Close() })
 		assert.Equal(t, http.StatusNoContent, resp.StatusCode)
 	})
+
+	t.Run("builds RootCAs from system pool plus configured bundle", func(t *testing.T) {
+		t.Parallel()
+		srv := httptest.NewTLSServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
+		t.Cleanup(srv.Close)
+
+		pemBytes := leafCertPEM(t, srv.TLS)
+		path := filepath.Join(t.TempDir(), "bundle.pem")
+		require.NoError(t, os.WriteFile(path, pemBytes, 0o600))
+
+		rt, err := newCustomCATransport(path)
+		require.NoError(t, err)
+
+		transport, ok := rt.(*http.Transport)
+		require.True(t, ok, "transport should be *http.Transport")
+		require.NotNil(t, transport.TLSClientConfig)
+		require.NotNil(t, transport.TLSClientConfig.RootCAs)
+
+		expectedPool, err := x509.SystemCertPool()
+		require.NoError(t, err)
+		require.True(t, expectedPool.AppendCertsFromPEM(pemBytes))
+		assert.True(t, transport.TLSClientConfig.RootCAs.Equal(expectedPool))
+	})
 }
 
 // TestNewClient_CACertFile verifies that [NewClient] surfaces an error from the
 // CA bundle loader rather than silently falling back to the default trust store
 // when [Config.CACertFile] is set but cannot be parsed, and that a valid bundle
-// is wired into both the IDP token client and the inner API client.
+// is wired into both the IDP token client and the final API transport.
 func TestNewClient_CACertFile(t *testing.T) {
 	t.Parallel()
 
@@ -99,12 +123,12 @@ func TestNewClient_CACertFile(t *testing.T) {
 		client, err := NewClient(conf)
 		require.NoError(t, err)
 		require.NotNil(t, client)
-		// No CACertFile means apiHTTPClient stays on http.DefaultTransport,
+		// No CACertFile means apiTransport stays on http.DefaultTransport,
 		// matching pre-change behavior.
-		assert.Nil(t, client.credentials.apiHTTPClient.Transport)
+		assert.Same(t, http.DefaultTransport, client.credentials.apiTransport)
 	})
 
-	t.Run("valid bundle wires custom transport into the API client", func(t *testing.T) {
+	t.Run("valid bundle wires custom transport into the API transport", func(t *testing.T) {
 		t.Parallel()
 		// Generate a real PEM bundle to satisfy the loader. The chain it certifies
 		// is unrelated to anything we dial here, the assertion is purely structural.
@@ -124,8 +148,8 @@ func TestNewClient_CACertFile(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, client)
 
-		apiTransport, ok := client.credentials.apiHTTPClient.Transport.(*http.Transport)
-		require.True(t, ok, "apiHTTPClient transport should be *http.Transport")
+		apiTransport, ok := client.credentials.apiTransport.(*http.Transport)
+		require.True(t, ok, "apiTransport should be *http.Transport")
 		require.NotNil(t, apiTransport.TLSClientConfig)
 		require.NotNil(t, apiTransport.TLSClientConfig.RootCAs,
 			"RootCAs must be set so Go's pure-Go x509 verifier (not the platform verifier) is used")
```

diliuskh:
Hey @nieomylnieja, sorry for the radio silence — got pulled into other stuff and only just circled back. Applied your patch on top of the branch in `fd00eef`.

One tiny deviation: golangci-lint's `unconvert` flagged `transport := http.RoundTripper(http.DefaultTransport)` (already typed `RoundTripper`), so I dropped to `transport := http.DefaultTransport`. Same semantics. Shout if you'd rather have the explicit cast back.

Verified locally — tests + lint green, smoke-tested end-to-end against Nobl9 on my corp laptop, TLS handshake works with the bundle and fails cleanly on a bogus path.

Thanks a ton for the review and for writing out the patch 🙏 ready whenever you are for another look.

nieomylnieja (approved):

nieomylnieja:
hey @diliuskh, happy to let you know that I've merged your PR, it is not yet officially released for the SDK, If you wish to use it directly with nobl9-go you can use the release candidate version [v0.131.0-rc3](https://github.com/nobl9/nobl9-go/releases/tag/v0.131.0-rc3)
it is however officially shipped with [sloctl v0.23.0](https://github.com/nobl9/sloctl/releases/tag/v0.23.0)
thanks again for the contribution ;)

PR #915, opened 2026-05-06
feat: add Reliability Rollup report variant

nikodemrafalski:
Adds v1alpha manifest support for Reliability Rollup reports so they can be defined and validated like the existing report variants.

## Summary

| Area | Behavior |
|---|---|
| Report type | `spec.reliabilityRollup` |
| Time frame | Rolling or calendar |
| Scope | Filters or custom hierarchy |
| Hierarchy | Folders and SLO refs |
| Depth limit | 8 folder levels |

`spec.filters` and `spec.reliabilityRollup.customHierarchy` are mutually exclusive. Existing report types still require filters.

## Release Notes

Added v1alpha manifest support for Reliability Rollup reports.


skrolikiewicz (commented):
## Code review — Reliability Rollup report variant (SDK)

Reviewed `feat/report-reliability-rollup`. `go build ./...` and `go test ./manifest/v1alpha/report/...` pass locally. This is a clean, thoroughly-tested addition — 24 validation test cases covering the filters/customHierarchy XOR rule, the depth cap, time-frame rules, and a pathological-depth case. Branch commit `cc39262` matches the pseudo-version n9 #20878 pins, so the two repos are in sync.

### Observations

**Recursive vs iterative hierarchy handling (minor, low severity).** `hierarchyDepthExceeds` is deliberately iterative with an explicit stack — the comment says this is to keep a pathological input from exhausting the goroutine stack. But the per-folder validation walk is genuinely recursive: `buildHierarchyFolderValidation` validates `children` via `folderValidator.Validate(child)`. The depth-check rule and the recursive `IncludeForEach` walk are two independent property rules on `customHierarchy`, so the recursive walk is not actually gated by the depth check — both execute. The 100k-depth test passes (Go grows the stack and real inputs are capped at 8), so this is fine in practice; just flagging the asymmetry, since the iterative depth check implies the recursion was already a known concern.

**Timezone validation is inline-duplicated.** The `time.LoadLocation` `timeZone` rule is copy-pasted across `validation_reliability_rollup.go`, `validation_slo_history.go`, and `validation_system_health_review.go`. This PR follows the existing pattern so it is not a regression — but a shared `timeZoneValidation` rule would be worth extracting at some point.

### Confirmed consistent with n9 #20878

- Depth cap of 8 (`MaxHierarchyDepth` here, `MaxReliabilityRollupHierarchyDepth` in n9) and the depth algorithm (root = depth 1, `> max` rejects) match exactly.
- The filters/customHierarchy XOR rule matches n9's `reliabilityRollupDefinitionValidation`. The implementations differ intentionally — the SDK keys off `filters != nil`, while n9 keys off `!Filters.IsEmpty()` because n9's `Normalize()` materializes a non-nil empty `Filters`.
- `HierarchyFolder` / `HierarchySLORef` field shapes round-trip cleanly through n9's `ParseV1AlphaCustomHierarchy` / `parseHierarchyToV1Alpha`.

No blockers — solid PR.

manifest/v1alpha/report/validation_reliability_rollup.go
```diff
@@ -0,0 +1,125 @@
...
+			Rules(govy.NewRule(func(f HierarchyFolder) error {
+				if len(f.Children) == 0 && len(f.SLOs) == 0 {
+					return errors.New("folder must contain at least one child folder or slo")
+				}
+				return nil
+			})),
+		govy.ForSlice(func(f HierarchyFolder) []HierarchyFolder { return f.Children }).
+			WithName("children").
+			RulesForEach(govy.NewRule(func(child HierarchyFolder) error { //nolint:gocritic
+				return folderValidator.Validate(child)
```
skrolikiewicz, line 90: `folderValidator.Validate(child)` recurses through the entire hierarchy. `hierarchyDepthExceeds` (line 60) was deliberately made iterative to avoid exhausting the goroutine stack — but it is a separate property rule from this `IncludeForEach` walk, so the depth check does not gate this recursion; both execute. The 100k-depth test passes (Go grows the stack, and real inputs cap at 8), so this is fine in practice — just flagging the iterative-vs-recursive asymmetry.

manifest/v1alpha/report/validation_reliability_rollup.go
```diff
@@ -0,0 +1,125 @@
...
+		Rules(rules.MutuallyExclusive(true, map[string]func(t ReliabilityRollupTimeFrame) any{
+			"rolling":  func(t ReliabilityRollupTimeFrame) any { return t.Rolling },
+			"calendar": func(t ReliabilityRollupTimeFrame) any { return t.Calendar },
+		})).
+		Include(govy.New[ReliabilityRollupTimeFrame](
+			govy.For(func(t ReliabilityRollupTimeFrame) string { return t.TimeZone }).
+				WithName("timeZone").
+				Required().
+				Rules(govy.NewRule(func(v string) error {
+					if _, err := time.LoadLocation(v); err != nil {
```
skrolikiewicz: This `time.LoadLocation` timeZone rule is copy-pasted in `validation_slo_history.go` and `validation_system_health_review.go` as well. It follows the existing pattern so it is not a regression — but a shared `timeZoneValidation` rule would be worth extracting at some point.

skrolikiewicz (commented):
Adding line-anchored notes for the two points from my earlier review.

skrolikiewicz:
### Post-review update

I pushed a commit to this branch — `51673bc` `fix(reports): relax RRR validation to match the web UI` — addressing two RRR bugs QA found on dev-demo-3:

- **Empty folders.** A `customHierarchy` folder with no children and no SLOs was rejected with `folder must contain at least one child folder or slo`. The web UI allows empty folders, so that rule was dropped. This supersedes the "folder validation looks correct" implication in my review above.
- **Unclear required-field error.** A report with neither `filters` nor `customHierarchy` reported only `spec.filters: property is required`. The spec rule now reports `spec.filters or spec.reliabilityRollup.customHierarchy is required`, and the generic `filters` `Required()` is scoped to non-reliability-rollup report types.

I also tagged **`v0.127.0-rc2.1`** at this commit so downstream repos (n9) can pin a tagged version instead of a branch pseudo-version — worth confirming with the branch owner, since it's a published version in the SDK namespace.

The two inline comments above (recursive per-folder validation walk; duplicated timezone rule) still stand.

skrolikiewicz (commented):
## Re-review — all prior review points addressed

Re-reviewed at branch HEAD `654e1b3` (now merged up to `main` via `514ba19`). `go build ./...` and `go test ./manifest/v1alpha/report/...` pass.

Status of everything raised across the earlier reviews:

| Item | Status |
|---|---|
| QA bug — empty `customHierarchy` folders rejected by sloctl | ✅ Fixed (`51673bc`). Rule removed; covered by `passes when nested folder is empty` and `passes with a deeply nested hierarchy of empty folders`. |
| QA bug — unclear error when neither `filters` nor `customHierarchy` is set | ✅ Fixed (`51673bc`). Now reports `spec.filters or spec.reliabilityRollup.customHierarchy is required`. |
| Recursive per-folder validation not gated by the depth check | ✅ Resolved (`cdb1199`). The depth rule and `IncludeForEach` are now a single `ForSlice` rule with `Cascade(CascadeModeStop)`, so the recursive walk is skipped once the depth check fails — covered by the new `stops customHierarchy validation when depth exceeds maximum` test. |
| `time.LoadLocation` timezone rule duplicated across 3 validators | ✅ Resolved (`2508e9c`). Extracted shared `timeZoneValidationRule` in `validation_time_zone.go`, now used by RRR / SLOHistory / SystemHealthReview. |
| Branch stale relative to `main` | ✅ `main` merged in (`514ba19`). |

No new issues in the follow-up commits — the `Cascade(CascadeModeStop)` change and the timezone extraction are both clean and test-covered.

Good to go from my side. Not approving formally — I've pushed commits to this branch, so a CODEOWNER review remains the right gate.

One coordination note (not a #915 blocker): the `v0.127.0-rc2.1` pre-release tag points at `51673bc` and is now 5 commits behind HEAD — it carries the QA fixes but not `cdb1199` / `2508e9c` / the `main` merge. Anything downstream pinning that tag (e.g. n9) would need a fresh tag to pick those up.

skrolikiewicz (approved):

PR #916, opened 2026-05-07
feat: add Dynatrace DQL SDK support

matkaras:
## Release Notes

Add SDK support for Dynatrace DQL SLO metric queries with optional `dql.interval` and validation that Nobl9-controlled time parameters stay out of the DQL query.

## Motivation

Dynatrace users could only model Metrics API `metricSelector` queries in manifests, so DQL-backed SLOs could not be defined as a first-class feature. Add DQL as a separate query shape in the SDK so all callers share one manifest contract and validation layer.

## Summary

Add `dql.query` and optional `dql.interval` to Dynatrace metric queries. Validate that exactly one of `metricSelector` or `dql` is configured, require `dql.interval` to be at least `15s` when provided, add Direct `platformToken` support, and reject DQL query-level `from:`, `to:`, `timeframe:`, `interval:`, `bins:`, and `shift:` through standard manifest validation.


manifest/v1alpha/slo/metrics_dynatrace.go
```diff
@@ -1,18 +1,98 @@
...
+)
+
+var dynatraceMetricQueryRequiredRule = govy.NewRule(func(d DynatraceMetric) error {
+	if !d.IsMetricSelectorConfiguration() && !d.IsDQLConfiguration() {
+		return errors.New("one of 'metricSelector' or 'dql' is required")
+	}
+	return nil
+}).WithErrorCode(rules.ErrorCodeRequired)
+
+var dynatraceMetricQueryMutuallyExclusiveRule = govy.NewRule(func(d DynatraceMetric) error {
```
nieomylnieja: there's already a rule for that: https://pkg.go.dev/github.com/nobl9/govy@v0.26.0/pkg/rules#MutuallyExclusive

If you set the first parameter to `true` it will also cover `dynatraceMetricQueryRequiredRule`

manifest/v1alpha/slo/metrics_dynatrace.go
```diff
@@ -54,10 +52,20 @@ func (d DynatraceMetric) QueryType() DynatraceMetricQueryType {
...
+				}
+				return nil
+			},
+			"dql": func(d DynatraceMetric) any {
+				if d.IsDQLConfiguration() {
+					return d.DQL
+				}
+				return nil
+			},
+		})),
```
nieomylnieja, line 61: is this correct? why not just directly assert the pointer? after all, since you defined them as pointers, an empty string should mean that the property was explicitly set to an empty string:

```go
	govy.For(govy.GetSelf[DynatraceMetric]()).
		Rules(rules.MutuallyExclusive(true, map[string]func(d DynatraceMetric) any{
			"metricSelector": func(d DynatraceMetric) any { return d.MetricSelector },
			"dql":            func(d DynatraceMetric) any { return d.DQL },
		})),
```

you can see examples of it in other places: https://github.com:nobl9/nobl9-go/blob/f84ff4fc4ff96f28e61bd299fb4a78e22ad4bebd/metrics_validation.go#L40-L54

nieomylnieja (changes requested):

nieomylnieja (approved):

nieomylnieja (approved):

PR #925, opened 2026-05-20
feat: PC-19870 allow alpha and beta release channels for Splunk Observability

bdw-nobl9:
Allow Splunk Observability data sources on both the `alpha` and `beta` release channels (Direct and Agent), instead of beta only. Splunk Observability stays available on `alpha` while Replay graduates onto `beta`.

## Behavior

| Spec | Before (beta-only) | After |
|---|---|---|
| Splunk Observability Direct | `beta` only | `alpha` or `beta` |
| Splunk Observability in alpha-enabled sources | dropped | kept |
| Splunk Observability Agent | gated to `beta` | `alpha` or `beta` |

## Edge Cases

- The agent gate mirrors the direct gate: `alpha` or `beta` only, `stable` rejected with a clear error. The agent example is generated on `beta`.
- Direct still rejects `stable` for Splunk Observability with a clear error.

### DIFF
--- a/manifest/v1alpha/examples/agent.go
+++ b/manifest/v1alpha/examples/agent.go
@@ -59,6 +59,7 @@ var betaChannelAgents = []v1alpha.DataSourceType{
 	v1alpha.Elasticsearch,
 	v1alpha.ThousandEyes,
 	// Support for Replay only in beta.
+	v1alpha.SplunkObservability,
 	v1alpha.SumoLogic,
 	v1alpha.Atlas,
 	v1alpha.Dash0,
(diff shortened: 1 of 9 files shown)


manifest/v1alpha/examples/agent.go
```diff
@@ -58,6 +58,8 @@ var betaChannelAgents = []v1alpha.DataSourceType{
 	// Support for Replay only in beta.
 	v1alpha.Elasticsearch,
 	v1alpha.ThousandEyes,
+	// Gated to alpha and beta release channels.
+	v1alpha.SplunkObservability,
```
matkaras:
```suggestion

```

matkaras (approved):

PR #928, opened 2026-05-22
fix: channel-key static fixture names

bdw-nobl9:
Persisted e2e fixtures were named by `(kind, type)` only, so flipping an example's release channel turned the next Apply into an UPDATE the backend rejects with `can't switch between release channels`. Suffix the fixture name and cache key with the channel for non-stable values so flips become a CREATE of a distinct object.

## Release Notes

Fix e2e test helper rejecting Apply when an example's release channel changes.

## Related PRs

- nobl9-go #926 "fix: set Splunk Observability agent example to beta channel" (merged 2026-05-21) - flip that exposed the issue
- terraform-provider-nobl9 #567 "feat: add Splunk Observability historical retrieval" (merged 2026-05-21) - consumer RC that surfaced the failure


bdw-nobl9:
Superseded by #929 (variant approach picked instead - splits Splunk Obs agent example into stable canonical + replay subvariant, which sidesteps the channel-switch backend rule without needing fixture-name channel keying). Closing as not needed.

PR #933, opened 2026-05-28
feat: Add SDK support for Alert Policy Analyzer

nobl9-adam-szymanski:
## Motivation

This enables API clients and MCP tools to run and inspect Alert Policy Analyzer workflows through the Go SDK.

## Summary

Added `AlertAnalysis().V1()` SDK methods for the Alert Policy Analyzer API:
- `StartAnalysis` starts an analysis run and returns analysisId.
- `GetAnalysis` returns analysis status, results, and optional time series.
- `RetryAnalysis` retries a failed analysis.
- `CalculateAlertPolicy` returns suggested alert policies for an SLO objective in a time range.

## Release Notes

Added Alert Policy Analyzer API which allows users to analyze Alert Policies and adjust their settings based on existing SLO data. The API is available under `sdk.Client.AlertAnalysis.V1`.

PR #941, opened 2026-06-10
feat: add Status Page endpoints

skrolikiewicz:
## Motivation

The Nobl9 platform serves an organization status page read API (components with computed statuses, disruptions), but the SDK had no support for it — consumers had to hand-roll HTTP clients and response models to reach it.

## Summary

Adds a `StatusPage()` endpoint group, mirroring the `AlertAnalysis()` pattern from #933:

- `client.StatusPage().V1().GetStatus(ctx)` — `GET /api/dashboards/v1/status-page/status`. Returns the component tree with each component's computed status (`operational` / `degradedPerformance` / `majorOutage`), no-signal flag, SLO references, any impacting disruption, and per-day disruption history.
- `client.StatusPage().V1().ListDisruptions(ctx, params)` — `GET /api/dashboards/v1/status-page/disruptions`. Optional `State` (`impacting`/`cleared`; empty returns both), `Limit` (server default 50, max 200), `Offset`.

Notes:

- Models mirror the server JSON exactly; names use the user-facing **disruption** terminology, consistent with the UI and [docs](https://docs.nobl9.com/organization-status/).
- `StatusComponent.Children` is self-referential by design (faithful tree); consumers that can't handle recursive schemas must flatten it themselves — documented on the type.
- Requires the caller's organization to have the Status Page feature; access control is enforced server-side.

## Related changes

None.

## Testing

- e2e tests in `tests/statuspage_v1_test.go` (per review, instead of httptest-based unit tests): `GetStatus` validates every component in the tree recursively (ID, name, status enum), `ListDisruptions` covers both `impacting` and `cleared` states with limit/total and `isCleared` consistency checks. The status page has no SDK API to create components, so the tests assert call success and response shape against whatever the e2e org (which has the feature enabled) serves.
- Manual check against an organization with the Status Page feature:

  ```go
  client, _ := sdk.DefaultClient()
  status, err := client.StatusPage().V1().GetStatus(ctx)
  disruptions, err := client.StatusPage().V1().ListDisruptions(ctx, statuspagev1.ListDisruptionsRequest{})
  ```

## Release Notes

Added Status Page endpoints:
- `client.StatusPage().V1().GetStatus()` returns the organization status page component tree with computed statuses and impacting disruptions
- `client.StatusPage().V1().ListDisruptions()` lists open or resolved disruptions


sdk/client_statuspage_v1_test.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+	defer srv.Close()
+
+	response, err := client.StatusPage().V1().ListDisruptions(
+		context.Background(),
+		statusPageV1.ListDisruptionsRequest{},
+	)
+	require.NoError(t, err)
+	assert.Empty(t, response.Disruptions)
+}
```
nieomylnieja: let's avoid writing these tests, prefer to write e2e in `tests` folder

nieomylnieja (changes requested):

sdk/client_statuspage_v1_test.go
skrolikiewicz: Done — removed the unit tests and added `tests/statuspage_v1_test.go` following the alertanalysis e2e pattern. The status page has no SDK API to create components, so the tests verify both calls succeed against the e2e org and validate the decoded shapes — component statuses recursively, disruption state consistency, and limit/total invariants. (Branch was squashed since.)

tests/statuspage_v1_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	statusPageV1 "github.com/nobl9/nobl9-go/sdk/endpoints/statuspage/v1"
+)
+
+func Test_StatusPage_V1_GetStatus(t *testing.T) {
+	t.Parallel()
+
+	response, err := tryExecuteRequest(t, func() (statusPageV1.GetStatusResponse, error) {
```
nieomylnieja: tryExecuteRequest only makes sense for tests where we apply some state and then the state is propagated asynchronously, this way we kind of wait until it is avialable
in this case, just call the endpoint directly, same for the next test

sdk/endpoints/statuspage/v1/response.go
```diff
@@ -0,0 +1,114 @@
+package v1
+
+import "time"
+
+//cspell:ignore getstatus listincidents getincident getcomponentdetails
```
nieomylnieja: let's stick to listing cspell exceptions in cspell.json

sdk/endpoints/statuspage/v1/response.go
```diff
@@ -0,0 +1,114 @@
...
+
+// The models in this file mirror the JSON served by the Nobl9 dashboards service
+// (n9 repo: dashboards/internal/handlers/getstatus.go, listincidents.go,
+// getincident.go, getcomponentdetails.go and dashboards/internal/repository/incidents.go).
+// Field names use the user-facing "disruption" terminology; JSON tags must match
+// the server exactly.
+
+// ComponentStatus is the computed status of a status page component.
+// Disruption severities share the same values, except "operational".
```
nieomylnieja:
```suggestion
// Disruption severities share the same values, except [ComponentStatusOperational].
```

sdk/endpoints/statuspage/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package statuspage provides access to Nobl9 Status Page APIs.
```
nieomylnieja: good place to reference some docs if we have any

nieomylnieja (changes requested):

tests/statuspage_v1_test.go
skrolikiewicz: Done in d6fd3e0e — both tests call the endpoints directly now.

sdk/endpoints/statuspage/v1/response.go
skrolikiewicz: The `cspell:ignore` line is already gone from the current head — it was removed together with the comment that needed it (the branch was squashed in the meantime). `statuspage` itself is listed in cspell.json.

sdk/endpoints/statuspage/v1/response.go
skrolikiewicz: Applied in d6fd3e0e.

sdk/endpoints/statuspage/doc.go
skrolikiewicz: Done in d6fd3e0e — linked the Organization status docs (https://docs.nobl9.com/organization-status/) from the package doc.

nieomylnieja (approved):

PR #945, opened 2026-06-12
feat: add limit support to users v2 list requests

nieomylnieja:
## Motivation

Allow `GetUsers` to pass a `limit` query parameter through the SDK.
Without it the API is limited to 100 results and there's no way to circumvent it.

## Related changes

https://github.com/nobl9/sloctl/pull/483

## Testing

Expanded users v2 tests to cover fetching all users and limiting results.

## Release Notes

`GetUsers` now supports a limit parameter.


sdk/endpoints/users/v2/request.go
```diff
@@ -2,10 +2,11 @@ package v2
...
 type GetUsersRequest struct {
-	IDs []string
+	IDs   []string
+	Limit uint
 }
 
 type getUsersRequest struct {
 	Phrase string
-	IDs    []string
+	GetUsersRequest
```
matkaras, line 11: question: what is reasoning to make it nested struct?  Can't just be inline fields?

nieomylnieja: because the underlying API is the superset, every parameter GetUsersRequest will have would be supported

matkaras: ah ok I see now 👍

matkaras (approved):

PR #946, opened 2026-06-13
feat: extend SDK API error fields

matkaras:
## Release Notes

SDK API errors can now expose optional status, detail, and metadata fields.

## Motivation
AgentCommander interactive command errors need JSON:API-style status and detail fields, but sdk.APIError only exposed title, code, and source. Add optional fields so API services can return richer errors without changing existing clients.

## Summary
Add optional status, detail, and meta fields to sdk.APIError. Add JSON compatibility coverage for minimal and extended error payloads.

## Related Changes
None.


sdk/http_errors.go
```diff
@@ -82,8 +82,14 @@ type APIError struct {
 	Title string `json:"title"`
 	// Code is an application-specific error code. It is optional.
 	Code string `json:"code,omitempty"`
+	// Status is the HTTP status code applicable to this error. It is optional.
+	Status string `json:"status,omitempty"`
```
nieomylnieja: we already have `Code`, which was intended to be this generic field for HTTP statuses, but not limited to them (e.g. govy error codes)

nieomylnieja (changes requested):

sdk/http_errors.go
matkaras: I was looking at https://jsonapi.org/format/#error-objects 
* code: an application-specific error code, expressed as a string value.
* status: the HTTP status code applicable to this problem, expressed as a string value. This SHOULD be provided.

skrolikiewicz: we don't have to (and shouldn't) strictly follow the jsonapi requirements

matkaras: We don't have to follow it, but it looks like a good fit, and it's a standard that we partially support. The title itself is not enough to provide the full context for query datasource endpoint. As this is a good candidate for a public API, I'd prefer not to return a custom envelope that could drift over time from the standard error that we return from the other endpoints.

Example of usage:

```
 HTTP/1.1 422 Unprocessable Entity
  Content-Type: application/json

  {
    "errors": [
      {
        "status": "422",
        "title": "Datasource command failed",
        "code": "datasource_command_failed",
        "detail": "SumoLogic query failed: failed to fetch Metrics: invalid status code returned
        for POST https://api.us2.sumologic.com/api/v1/metricsQueries - [status code: 401, want: <
        300, body: Credential could not be verified.]",
        "meta": {
          "retryable": false
        }
      }
    ]
```

skrolikiewicz: The HTTP response has one status, but errors is a list. A status on each error looks like it should match the response status, but nothing stops two errors from having different values - and then which one is right? We already have code for the app-level error, and the HTTP status is on the response itself, so status just adds confusion.

On `meta` - it was actually in the first version of the spec. Not sure why it didn't make it in, probably just wasn't needed yet.

Image

matkaras: I don't insist on the status being present, and I can remove it. However, this doesn't mean that a status code for multiple errors doesn't exist today; it's just hidden.

nieomylnieja: @matkaras, the status code is already present in the wrapping HTTPError in the SDK, please consider it in the context of that wrapping error, not just the APIError itself

matkaras: For me, the APIError reflects it better and the status isn't that important after all.

nieomylnieja (approved):

PR #947, opened 2026-06-14
feat: add project scope setting to report filters

nikodemrafalski:
## Motivation

PC-19834 adds report project scope to Nobl9 Report manifests, so SDK users can represent reports that cover all current and future projects without enumerating project names.

## Summary

Added `ProjectScope` constants and `projectScope` JSON support to Report filters. `projectScope: all` now satisfies Report filter validation on its own, while unknown values and `all` combined with explicit projects, services, or SLOs are rejected.

Preserved backward compatibility for existing Report manifests that omit `projectScope`: they still validate with explicit resources and marshal without emitting the new field unless it is set.

## Testing

Added manifest validation coverage for `projectScope: all`, invalid project scope values, and invalid combinations with explicit resources. Added JSON round-trip coverage for new all-project filters and legacy filters that omit `projectScope`.

E2E tests were not added because this change is limited to manifest encoding and validation, not SDK client behavior against the Nobl9 platform API.

## Release Notes

Added `projectScope` to Report manifest filters.

### DIFF
--- a/manifest/v1alpha/report/validation.go
+++ b/manifest/v1alpha/report/validation.go
@@ -93,12 +93,34 @@ var specValidation = govy.New[Spec](
 
 var filtersValidation = govy.New[Filters](
 	govy.For(govy.GetSelf[Filters]()).
+		When(
+			func(f Filters) bool { return f.ProjectScope == ProjectScopeAll },
+			govy.WhenDescription("projectScope is set to all"),
+		).
 		Rules(govy.NewRule(func(f Filters) error {
-			if len(f.Projects) == 0 && len(f.Services) == 0 && len(f.SLOs) == 0 {
-				return errors.New("at least one of the following fields is required: projects, services, slos")
+			if len(f.Projects) > 0 || len(f.Services) > 0 || len(f.SLOs) > 0 {
+				return errors.New("projectScope=all cannot be combined with projects, services, or slos")
 			}
 			return nil
 		})),
+	govy.For(govy.GetSelf[Filters]()).
+		When(
+			func(f Filters) bool { return f.ProjectScope == ProjectScopeSelected || f.ProjectScope == "" },
+			govy.WhenDescription("projectScope is selected or unspecified"),
+		).
+		Rules(govy.NewRule(func(f Filters) error {
+			if len(f.Projects) == 0 && len(f.Services) == 0 && len(f.SLOs) == 0 {
+				return errors.New("at least one of the following fields is required: projectScope=all, projects, services, slos")
+			}
+			return nil
+		})),
+	govy.For(func(f Filters) ProjectScope { return f.ProjectScope }).
+		WithName("projectScope").
+		When(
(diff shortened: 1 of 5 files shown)


manifest/v1alpha/report/report_test.go
```diff
@@ -0,0 +1,63 @@
...
+	require.NotNil(t, report.Spec.Filters)
+	assert.Empty(t, report.Spec.Filters.ProjectScope)
+	assert.Equal(t, []string{"project"}, report.Spec.Filters.Projects)
+	assert.Nil(t, validate(report))
+
+	data, err := json.Marshal(report)
+	require.NoError(t, err)
+	assert.NotContains(t, string(data), "projectScope")
+	assert.Contains(t, string(data), `"projects":["project"]`)
+}
```
nieomylnieja: I'm pretty sure we don't need these tests
If you want some flavour testing, extend examples with this new variant

manifest/v1alpha/report/validation.go
```diff
@@ -94,11 +94,21 @@ var specValidation = govy.New[Spec](
...
 			}
 			return nil
 		})),
+	govy.For(func(f Filters) ProjectScope { return f.ProjectScope }).
+		WithName("projectScope").
+		When(
+			func(f Filters) bool { return f.ProjectScope != "" },
+			govy.WhenDescription("projectScope is set"),
+		).
+		Rules(rules.OneOf(ProjectScopeSelected, ProjectScopeAll)),
```
nieomylnieja, line 120: wouldn't it be cleaner to have two validation branches with When condition, one when projectScope is set to all, and the other one when projectScope is empty or set to selected
then you could have cleaner rules for both of them, wdyt?

nieomylnieja (changes requested):

skrolikiewicz (approved):
The rules cover the combinations without overlapping into double-errors, an unset scope correctly defaults to `selected`, and allowing `all` alongside labels-only is the right call. Nice test coverage across all the branches, and the example + golden YAML keep things in sync.

manifest/v1alpha/report/validation.go
```diff
@@ -93,12 +93,34 @@ var specValidation = govy.New[Spec](
...
+				return errors.New("at least one of the following fields is required: projectScope=all, projects, services, slos")
 			}
 			return nil
 		})),
+	govy.For(func(f Filters) ProjectScope { return f.ProjectScope }).
+		WithName("projectScope").
+		When(
+			func(f Filters) bool { return f.ProjectScope != "" },
+			govy.WhenDescription("projectScope is set"),
+		).
```
nieomylnieja: `OmitEmpty` will handle this

nieomylnieja (approved):

manifest/v1alpha/report/validation.go
nikodemrafalski: thx

PR #952, opened 2026-06-19
feat: add data source query endpoint

matkaras:
## Release Notes

Add SDK support for querying Nobl9 data sources with `Client.DataSource().V2().Query()`.

## Motivation

The MCP server needs to query Nobl9 data sources through a typed SDK endpoint instead of manually constructing AgentCommander HTTP requests. Adding this SDK endpoint centralizes the request shape and preserves standard SDK HTTP error handling.

## Summary

Add the `DataSource` SDK entrypoint with a v2 `Query` endpoint for time-series data source queries.
Model public `QueryRequest` and `QueryResponse` types while mapping internally to `POST /api/agentcommander/v2/commands/timeseries/execute`.
Cover successful request forwarding, response decoding, and `sdk.APIErrors` propagation.


sdk/endpoints/datasource/v2/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package v2 provides access to Nobl9 data source APIs.
+package v2
```
nieomylnieja: you want it in v2, the version here is for the SDK, not for the underlying API
so if we would want to modify the SDK contract we would then have v2 on top of v1

sdk/endpoints/datasource/v2/request.go
```diff
@@ -0,0 +1,23 @@
...
+
+type QueryRequest struct {
+	DataSource v1alphaSLO.MetricSourceSpec `json:"datasource"`
+	Query      Query                       `json:"query"`
+	TimeRange  TimeRange                   `json:"timeRange"`
+}
+
+type Query struct {
+	RawMetric    *v1alphaSLO.RawMetricSpec    `json:"rawMetric,omitempty"`
+	CountMetrics *v1alphaSLO.CountMetricsSpec `json:"countMetrics,omitempty"`
```
nieomylnieja: remove json tags from Query and QueryRequest, it makes it a bit clearer that they are not used as the actual payload

nieomylnieja (changes requested):

nieomylnieja (approved):

PR #960, opened 2026-07-08
feat: Allow to configure no data anomaly for 0 values

nobl9-adam-szymanski:
## Motivation

This change adds SDK support for that field without changing existing behavior. The new field follows the existing optional boolean pattern so the SDK can distinguish between:
- field unset, which omits it from serialized YAML/JSON
- explicit `true`
- explicit `false`

That preserves the current default behavior when the field is not provided. This change is backward compatible, no breaking change.

## Summary

Added `TreatZeroAsNoData *bool` to `manifest/v1alpha/slo.AnomalyConfigNoData`.

Updated manifest/unit coverage to verify the expected serialization behavior:
- unset is omitted
- `true` is serialized
- `false` is serialized

## Testing

```yaml
anomalyConfig:
  noData:
    alertMethods:
    - name: slack-notification
      project: default
    alertAfter: 1h
    treatZeroAsNoData: true
```

## Release Notes

The SDK now exposes `anomalyConfig.noData.treatZeroAsNoData`, which controls whether zero values should be treated as no-data for SLO anomaly detection. When enabled, zero datapoints are interpreted as missing data and can trigger no-data anomaly behavior. The new field is optional, so can be omitted or set explicitly to true or false.


manifest/v1alpha/examples/slo_variants.go
```diff
@@ -182,7 +182,8 @@ func exampleAnomalyConfig() *v1alphaSLO.AnomalyConfig {
 					Project: sdk.DefaultProject,
 				},
 			},
-			AlertAfter: ptr("1h"),
+			AlertAfter:        ptr("1h"),
+			TreatZeroAsNoData: ptr(true),
```
nuusk:
```suggestion
			TreatZeroAsNoData: ptr(false),
```

this is in the shared SLO example helper. since the default is actually `false` i think we should make this `false` as well

nuusk (changes requested):
I think `false` makes more sense in the examples since the default is `false` in our codebase

manifest/v1alpha/examples/slo_variants.go
nobl9-adam-szymanski: If the example is shared/generic, the best value is neither true nor false:

- omit the field entirely

Because:
- false is default
- omission is intentionally equivalent to false
- explicit false makes the example noisier and less idiomatic

So the options are:
- generic helper: remove TreatZeroAsNoData
- feature-demonstration example: keep `TreatZeroAsNoData: ptr(true)`

I think `ptr(false)` is the weakest option of the three. So, or hide it in example or change it to `ptr(true` is the option worth to show in example I think.

nuusk: treatZeroAsNoData:true requires very specific scenarios to make sense. in most cases, people will have it as `false`, that's why having true in the examples is awkward to me. but since you already thought this through, I'm ok with whatever decision you make here

nuusk (approved):

manifest/v1alpha/examples/slo_variants.go
nobl9-adam-szymanski: Ok, I changed it to false. My second thought is that in most cases people will omit this or set it to false explicitly, so we still show this use case. We still show that field exist. Thanks for discussion.

PR #961, opened 2026-07-08
feat: add Replay annotation category

bdw-nobl9:
## Release Notes

Add SDK support for the `Replay` annotation category and its computed `spec.replay` facts.

## Summary

| Change | Detail |
|---|---|
| `Replay` category | Added to the enum and to `userCategories` |
| `spec.replay` | New `annotation.Replay` type: `periodStart`, `periodEnd`, `elapsedTimeSeconds`, tagged `nobl9:"computed"` |
| Description validation | Optional for `Replay`, required for every other category; the length bound still applies |
| v2 GET response | Carries `spec.replay` through from the annotations model |

## Motivation

The platform creates an annotation when a Replay finishes on an SLO. Without the category
in the SDK, the public annotations API, sloctl, and RBAC checks cannot parse those rows,
and the recalculated window and processing time have nowhere to surface.

## Why `Replay` is a user category

It sits in `userCategories` because the annotation carries the real requester as its
author and is listed and edited like a user annotation, so "Created by" matches
"Triggered by". Membership also makes `Replay` a valid `spec.category` value, without
which editing a Replay annotation fails validation. The platform, not the user, creates
it; n9 rejects a user-submitted `category: Replay` with a 400.

Consumers classify a category as system by its absence from the user list (n9 does
`!slices.Contains(GetUserCategories(), ...)`), so a category in neither list would silently
read as system; `TestCategory_EveryCategoryIsEitherSystemOrUser` guards the split.

## Edge cases

- `nobl9:"computed"` means a GET returns `spec.replay`. It does not mean an Apply strips
  it: `sdk.RemoveComputedFieldsFromObjects` is opt-in and neither `objects/v1.Apply` nor
  `objects/v2.Apply` calls it. Submitting the field is harmless, because the platform never
  derives the stored facts from a request.
- `elapsedTimeSeconds` is omitted when unset, and the whole block is omitted when absent.

## Related

Pairs with nobl9/n9#21247 (feat: create Replay annotation when a replay completes). Merge and
tag this first, then bump the n9 pin off the pre-merge pseudo-version.
Documentation: nobl9/documentation#1929 (docs: add Replay annotation category to annotations API reference).

## Testing

| Scenario | Expected |
|---|---|
| `Replay` category membership | In `GetUserCategories()`, not in `GetSystemCategories()`; every value appears in exactly one list |
| Facts serialization | Marshals under `spec.replay` only; marshal-unmarshal-marshal is stable |
| `RemoveComputedFieldsFromObjects` | Drops `spec.replay` when a caller invokes it |


manifest/v1alpha/annotation/category.go
```diff
@@ -33,14 +37,21 @@ var systemCategories = []Category{
...
 func GetSystemCategories() []Category {
 	return slices.Clone(systemCategories)
 }
 
-// GetUserCategories returns all annotation [Category] generated by the users.
+// GetUserCategories returns all annotation [Category] owned by users.
+// Users can edit and delete annotations in these categories, but not
+// necessarily create them: [CategoryReplay] annotations are created by the
+// Nobl9 platform when a replay completes, and users can only edit their
+// description. The API rejects attempts to create a Replay annotation.
```
matkaras, line 54: nitpick: I'd add this description here https://github.com/nobl9/nobl9-go/pull/961/changes#diff-9831e0e2d8f9cf9e450ed86813bfe658842fc3ec3bae4303f8069cc2dd0bcccaR40

manifest/v1alpha/annotation/validation.go
```diff
@@ -76,6 +77,16 @@ func getSpecValidation(includeUserCategoryRules bool) govy.Validator[Spec] {
...
 
+// Replay descriptions are optional user notes; other categories require a description.
+var descriptionRequiredForNonReplayValidation = govy.New[Spec](
+	govy.For(func(s Spec) string { return s.Description }).
+		WithName("description").
+		Required(),
+).When(
+	func(s Spec) bool { return s.Category != CategoryReplay },
+	govy.WhenDescription("category is not Replay"),
+)
```
matkaras, line 88: question: there will be no description for Replay annotations?

manifest/v1alpha/annotation/category.go
```diff
@@ -33,14 +37,21 @@ var systemCategories = []Category{
...
+// GetSystemCategories returns all annotation [Category] owned and managed
+// exclusively by the Nobl9 platform. Users cannot create or modify
+// annotations in these categories.
 func GetSystemCategories() []Category {
 	return slices.Clone(systemCategories)
 }
 
-// GetUserCategories returns all annotation [Category] generated by the users.
+// GetUserCategories returns all annotation [Category] owned by users.
+// Users can edit and delete annotations in these categories, but not
```
matkaras, line 51: question: can use delete Replay annotations?

manifest/v1alpha/annotation/validation.go
bdw-nobl9: No by default - but replay computed facts will carry the replay-specific information - the user will be able to add description as additional note containing context of a given replay. And even without description - UI will render the non-editable part (containing replay info like _Elapsed time_). Same for the returned annotation through API.

manifest/v1alpha/annotation/category.go
bdw-nobl9: Yes. Users with the regular Annotation Delete permission for the project can delete Replay annotations. Deletion is not restricted to the user who triggered the Replay.

manifest/v1alpha/annotation/category.go
bdw-nobl9: Fair, the list is where people actually trip over it. Added a short note on `CategoryReplay` in `userCategories` and kept the permission details in the `GetUserCategories` doc only. While there I fixed the `GetSystemCategories` doc, which said users cannot modify system annotations; they can delete them with the system-annotation delete permission, so the comment now says create/edit only: `b1269e4`. Shout if you want it worded differently.

manifest/v1alpha/annotation/validation.go
bdw-nobl9: Some numbers to back that up. On one of our dev environments today: 265 Replay annotations, 264 with an empty description and 1 where someone typed a note by hand, and all 265 carry `spec.replay` with `periodStart`, `periodEnd` and `elapsedTimeSeconds`. The platform creates them with an empty description on purpose; the replay facts are the content. On this side the rule is `descriptionRequiredForNonReplayValidation` in `validation.go:81-88`, and the length bound still applies to Replay. `TestValidate_Spec_Description` covers required for non-Replay, optional for Replay, and the length bound on Replay. Fine to resolve this one?

manifest/v1alpha/annotation/category.go
bdw-nobl9: Checked the platform side to be sure. Delete is permission-only: the delete path picks `annotation` or `system-annotation` as the resource based on whether the category is in `GetUserCategories()`, checks project access for delete, and never looks at ownership. So Replay being in `userCategories` here is exactly what puts it under the regular annotation delete permission (Org Admin, Org Responder, Project Owner/Editor/Responder in the default roles). Edit is stricter: only the user who requested the replay can edit, and only with the annotation edit permission on the project. The `GetUserCategories` doc now says this explicitly: `b1269e4`. Resolve?

PR #983, opened 2026-08-26
feat: add SLO list pagination and sorting

nieomylnieja:
## Motivation

The SLO GET client could only request the complete filtered result set and could not select a stable sort order.

## Summary

Added optional nested pagination and sorting objects to `GetSLOsRequest`. `go-enum` generates the supported sort values, while the query-key constants remain package-private. One govy validator owns the request rules and consumes the generated value lists. `GetV1alphaSLOs` calls `Validate()` before it encodes the nested query parameters. Omitting pagination remains unbounded.

## Testing

No SDK unit tests were added. API-backed E2E coverage exercises all eight supported sort combinations. It compares paginated slices with unpaginated results. It asserts that invalid raw query parameters get HTTP 400 responses. The E2E package passed a race-enabled compilation, and the repository generation check reproduced the enum output. Runtime execution requires an environment with the corresponding API behavior.

## Release Notes

The Objects V1 SDK can now paginate and sort `GetV1alphaSLOs` requests.


sdk/endpoints/objects/v1/request.go
```diff
@@ -17,11 +19,47 @@ type GetServicesRequest struct {
...
 type GetSLOsRequest struct {
-	Project  string
-	Names    []string
-	Labels   v1alpha.Labels
-	Services []string
+	Project    string
+	Names      []string
+	Labels     v1alpha.Labels
+	Services   []string
+	Pagination *GetSLOsPagination `form:"pagination"`
```
nikodemrafalski, line 29: **Blocking — this is an undeclared source-compatibility break.** Growing the exported `GetSLOsRequest` from four fields to six breaks existing downstream unkeyed literals: a literal that compiles against the base now fails against this head with `too few values in struct literal`. The repository permits pre-v1 minor breaks, but its PR template requires them to be listed under `## Breaking Changes`. Please either preserve source compatibility with a new request type/method, or explicitly document the break and migration to keyed fields.

sdk/endpoints/objects/v1/get_slos_request_test.go
```diff
@@ -0,0 +1,147 @@
...
+					PropertyPath:  "pagination.offset",
+					Code:          rules.ErrorCodeGreaterThanOrEqualTo,
+					ValidatorName: "Get SLOs request",
+				},
+			},
+		},
+		{
+			name: "offset above maximum",
+			request: GetSLOsRequest{
+				Pagination: &GetSLOsPagination{Limit: 1, Offset: maxGetSLOsOffset + 1},
```
nikodemrafalski: **Medium — this new test does not compile on 32-bit architectures.** `maxGetSLOsOffset + 1` is the untyped constant `2147483648`, which overflows the `int` `Offset` field on `GOARCH=386`. I confirmed the base package test binary cross-compiles, while this head fails at this line; the production package itself still cross-compiles. Please gate this above-maximum case on `strconv.IntSize`, use an architecture-independent validator test, or use a fixed-width field if values above 32-bit must be representable everywhere.

nikodemrafalski (commented):
Two confirmed compatibility findings from the current PR head.

sdk/endpoints/objects/v1/request.go
nieomylnieja: let's be real here, our API is the only decoding user of this and it doesn't break it, I did however add the rest of `form` tags to be clear here

sdk/endpoints/objects/v1/get_slos_request_test.go
nieomylnieja: yeah, I don't think we need to worry or care about 32-bit archs ;)
but coming to think about it, I think this particular validation is just not needed in any case, it's over the top

nikodemrafalski (approved):

PR #995, opened 2026-09-17
feat: add IsComposite to Replay list items PC-19980

bdw-nobl9:
## Motivation

With composite Replays appearing alongside regular Replays, clients need to identify which kind of SLO is being replayed. `ReplayListItem.IsComposite` preserves the distinction between a regular SLO (`false`) and an unknown value when the platform does not return the field yet or the caller cannot view the SLO (`nil`).

## Related changes

Consumed by [sloctl #549 (Replay list composite identification)](https://github.com/nobl9/sloctl/pull/549).

## Testing

Against a Nobl9 environment that already returns `isComposite`:

```sh
GOFLAGS='-run=^Test_Replay_V1$' make test/e2e
```

| Scenario | Expected |
|---|---|
| Queue a Replay for a regular SLO, then list | Its item has `IsComposite` set to `false` |
| Decode a list item without `isComposite` | `IsComposite` is `nil` |

## Release Notes

Replay lists now include composite SLO Replays when supported by the platform.


sdk/endpoints/replay/v1/response_test.go
```diff
@@ -24,6 +24,28 @@ func TestReplayListItemUnmarshal(t *testing.T) {
...
+		"unknown":       {json: `{"project":"p","status":"queued"}`, expected: nil},
+	} {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var item ReplayListItem
+			require.NoError(t, json.Unmarshal([]byte(test.json), &item))
+			assert.Equal(t, test.expected, item.IsComposite)
+		})
+	}
```
nieomylnieja: useless test, e2e already covers this

nieomylnieja (changes requested):

sdk/endpoints/replay/v1/response_test.go
bdw-nobl9: Removed the unit test and tightened the E2E assertion to require `isComposite` to be present and false for the regular SLO. Previously, nil silently skipped the check. Replay E2E passed on dev-grey.

nieomylnieja (approved):

